PR #75, opened 2013-11-18
Pass -Zgzip to dpkg-deb for backwards compatibility

moloch--:
Quick fix for issue https://github.com/DHowett/theos/issues/69


PaulBGD:
Still fixes it, still not pulled. Thanks!

kirb:
I would advise against using gzip. The actual problem here is that the dpkg version on the build machine is newer than the dpkg version in Telesphoreo. The latest dpkg supports (in order of compression ratio provided): gzip, bzip2, lzma, and xz. Telesphoreo’s dpkg build supports all of those except for xz. Therefore, the next best choice is lzma.

The flag `-Zlzma` was added some time ago in kirb/theos, which is now merged into this repo, so this PR can be closed.

ThePantsThief:
So are we all supposed to manually change it to gzip until Telesphoreo updates dpkg…?

kirb:
Telesphoreo dpkg supports LZMA, and this is the default.

ThePantsThief:
Oh. Do you have any idea why I'm getting an unsupported compression type error about it when I try to build a deb then?

--
Tanner Bennett

PaulBGD:
I'm still getting it too, this PR still fixes it.

DHowett:
That sounds like it's happening at build time and will have no bearing on
how it performs on-device.

Fwiw, this is precisely why I didn't want to set a default; reaching the
best supported feature set between commonly available dpkg versions is
fairly painful :p

bzip2 is likely to be the most widely supported version apart from gzip,
and it provides better ratios.

ThePantsThief:
Yeah, exactly haha. Does anyone know off the top of their head where to get a version for OS X that works with this?

--
Tanner Bennett

DHowett:
macports fairly expertly packages a wide range of updated versions of cool
things. This includes dpkg. :)

ThePantsThief:
Thanks G :)

--
Tanner Bennett

jackiossy:
How to fix it ?

PR #76, opened 2013-12-01
Add support for building on Windows (Cygwin).

coolstar:
Add support for building iOS apps from Windows :) The toolchain will be available very soon when I can package it up and make an installer for it. So far it seems to work fine, and the resulting binaries work perfectly :)

It is based on LLVM 3.3 too, so it works with ARC, Blocks, Literals, and the iOS 5, 6, and 7 SDKs.

Let me know what your thoughts are.

(No Anti-Windows comments please!)


kirb:
This branch was merged a while ago in kirb/theos (which is now merged into this repo).

PR #104, opened 2014-08-04
Update linux target to allow a newer toolchain based on LLVM+Clang.

coolstar:
This is actually identical to the target used on Windows as the toolchain is placed in $(THEOS)/toolchain/linux/iphone


kirb:
This has been implemented by @angelXwind in 26269dc.

PR #106, opened 2014-08-06
C function Logos directives

uroboro:
Added %hookf and %origf directives for hooking C functions while supporting groups and using a variant of %orig.
%hookf takes a variadic number of arguments: the function return type, the function name, the function argument list. It generates the original function function pointer from these values and the current group.
%hookf takes a variadic number to pass to the original function, which is calculated from saving the current function name in %hookf and the current group.


Tyilo:
Maybe add support for other functions than `MSHookFunction`? (eg. [rd_route](https://github.com/rodionovd/rd_route), [mach_override](https://github.com/rentzsch/mach_override))

Both of the functions they provide takes the same arguments in the same order as `MSHookFunction`, so you could just use `#define MSHookFunction rd_route`.
Maybe add a config option instead: `%config(hookf=rd_route)`?

ryanb93:
Would be great to see this merged soon.

rossistboss:
Why has this not been merged? Is there another way to hook C functions or something?

youmingtaiziback:
Cool feature

PR #115, opened 2015-02-13
Refer to protocols with @protocol

### LINKED ISSUE #114: %subclass with protocols crashes in the constructor
In CCLoader I have %subclass that implements two protocols, a custom one and one defined in SpringBoardUIServices. That subclass (sometimes) crashes in the constructor function. I'm using the latest theos. Here's the crash log, and the %subclass is also in that project: https://github.com/JonasGessner/CCLoader/issues/12

JonasGessner:
Fixes #114.

### DIFF
--- a/bin/lib/Logos/Generator/MobileSubstrate/Subclass.pm
+++ b/bin/lib/Logos/Generator/MobileSubstrate/Subclass.pm
@@ -23,7 +23,7 @@ sub initializers {
 	}
 	# </ivars>
 	foreach(keys %{$class->protocols}) {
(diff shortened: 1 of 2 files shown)


DHowett:
This only works with linkable protocols, which is infeasible for all use cases. It would probably be better to fix the problem in #114 with conditional initialization, or for Logos to emit code that captures/propagates errors generated due to Nil protocols.

kirb:
As this is pretty old and can’t be pulled without some work (https://github.com/theos/theos/issues/114#issuecomment-183460233), I’ll be closing this.

PR #158, opened 2016-05-17
[Logos] Fixed nil class spam

Stilldabomb:
Fixes constant spam to the syslog when a class cannot be found.

No

`[Tweak: Tweak.xm:<line>] ERROR: logos: nil class <class>`

Any other comments?
-------------------
Only stops error from being shown if binary is built in production (non-debug) mode.

### DIFF
--- a/bin/lib/Logos/Generator/MobileSubstrate/Method.pm
+++ b/bin/lib/Logos/Generator/MobileSubstrate/Method.pm
@@ -84,7 +84,9 @@ sub initializers {
 		$r .= "if (".$classvar.") {";
 		$r .=   "MSHookMessageEx(".$classvar.", ".$self->selectorRef($method->selector).", (IMP)&".$self->newFunctionName($method).", (IMP*)&".$self->originalFunctionName($method).");";
 		$r .= "} else {";
+		$r .= "#ifdef __DEBUG__";
 		$r .=   "HBLogError(@\"logos: nil class %s\", \"".$method->class->name."\");";
+		$r .= "#endif";
 		$r .= "}";
 	} else {
 		my $r = "";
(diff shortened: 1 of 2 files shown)


ahyattdev:
What is the benefit to having that log message?

This message makes the syslog from crashlogs hard to read. I will be glad to see it gone.

Stilldabomb:
@ahyattdev it allows for you to easily see if a class you are hooking is incorrect, which is useful for debugging if something isn't working correctly.

kirb:
Discussed this with @Stilldabomb; outcome is that we *may* look into making this a theosrc preference, but moreso that we believe this is really more of a flaw on the side of the projects that cause these logs.

Before this feature was added, these logs would have appeared anyway, because Substrate logs “nil class argument”. The idea of this was to have the same log message, but with the class name (it’s impossible for Substrate to determine the class name – all it gets is `Nil` – but Logos can).

These logs appearing should be considered as a bug in the tweak, even if it’s intentional. Tweaks should be using `%group`s, and then only initialising the group if it is sure that the classes/methods in question exist. It ensures the intent of your code is obvious; that nobody else reading the code (or yourself – don’t assume what you know about your code now will be remembered when you next revisit it) will get confused or make the wrong judgements on what the code is doing. 

@ahyattdev, this is rather vital for debugging. There’s no use in the code knowing something has gone wrong, but not telling you because you don’t like seeing the messages. You really don’t want to create more situations where you spend hours debugging an issue just to find at the end that it was because of something so obvious. I would much rather a ginormous syslog with only a few useful lines than a short syslog lacking vital information.

If you’re taking this stance because some other tweak logs a heap of these, urge its developer to read this thread.

kirb:
Since there hasn’t been any activity for some time (and because merging this branch will now cause a merge conflict), I‘ll be closing this. Reopen if you still think this should be merged.

PR #190, opened 2016-08-13
[Templates] Use /usr/local instead of /usr for tool and library templates

mstg:
This pull request changes the install path of the tool and library templates to point to `/usr/local/bin` and `/usr/local/lib`.

This is useful so we can separate system and user installed binaries/libraries. It also encourages a use of more modern convention for install locations.

No

No

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** macOS 10.12


kirb:
This was moved to [theos/nictemplates](https://github.com/theos/nicTemplates/pull/3), which will replace the templates/ dir soon.

PR #204, opened 2016-09-20
Workaround for WSL

DuIslingr:
## What does this implement/fix? Explain your changes.

This allows nic.pl to function on WSL(Windows SubSystem for Linux)
## Does this close any currently open issues?

Not that I know of.
## Any relevant logs, error output, etc?

Nope
## Any other comments?

Nope
## Where has this been tested?

**Operating System:** Windows 10 build 14926 (Should work fine on Anniversary 1607)

**Platform:** WSL Upgraded to Xenial 16.04.1 (Should also work on default 14.04)

**Target Platform:** iOS

**Toolchain Version:** AngelXWind hosted Linux Toolchain. Latest? Dunno the exact version number.

**SDK Version:** 9.2

### DIFF
--- a/bin/nic.pl
+++ b/bin/nic.pl
@@ -22,7 +22,15 @@ use NIC::Formats::NICTar;
 use NIC::NICType;
 
 # Workaround to allow Find to function on WSL. Currently on WSL without this line nic.pl will fail to find templates.
-$File::Find::dont_use_nlink=1;
+if ($^O eq "linux") {
+open(FILE,"/proc/version");
+if (grep{/Microsoft/} <FILE>){
+$File::Find::dont_use_nlink = 1;
+}else{
+$File::Find::dont_use_nlink = 0;
+}
+close FILE;
+}
 
 our $savedStdout = *STDOUT;
 
--- a/bin/nic.pl
+++ b/bin/nic.pl
@@ -21,6 +21,7 @@ use NIC::Bridge::Context (PROMPT => \&nicPrompt);
 use NIC::Formats::NICTar;
 use NIC::NICType;
 
+# Workaround to allow Find to function on WSL. Currently on WSL without this line nic.pl will fail to find templates.
 $File::Find::dont_use_nlink=1;
 
 our $savedStdout = *STDOUT;
--- a/bin/nic.pl
+++ b/bin/nic.pl
@@ -22,14 +22,16 @@ use NIC::Formats::NICTar;
 use NIC::NICType;
 
 # Workaround to allow Find to function on WSL. Currently on WSL without this line nic.pl will fail to find templates.
-if ($^O eq "linux") {
-open(FILE,"/proc/version");
-if (grep{/Microsoft/} <FILE>){
-$File::Find::dont_use_nlink = 1;
-}else{
-$File::Find::dont_use_nlink = 0;
-}
-close FILE;
+if ($ ^ O eq "linux") {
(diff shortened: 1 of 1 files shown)


DHowett:
Would you mind adding a comment explaining why it's necessary (in the
file), and fixing the spacing? :)

-- Dustin L. Howett

DuIslingr:
Currently on WSL Perl Find does not function as it is supposed to. This is a workaround that allows Perl to find the files, otherwise NIC.PL fails saying it cannot find any templates. I suppose it could be added to the other pl files that make use of Find if needed.

mstg:
@DuIslingr You have to set this WSL only somehow. Perl is not able to detect the WSL NTFS filesystem (which nlink has to be disabled for) and this problem occures. Be aware that this option may be dangerous as it will both slow down the Find process and follow symlinks. I suggest reading this: http://search.cpan.org/~rjbs/perl-5.24.0/ext/File-Find/lib/File/Find.pm

What do you guys think? @kirb @DHowett

kirb:
I think it'll be fine for now. I need to work on this code anyway to add support for `vendor/templates`, so might as well have a solution for now and I can implement a better one later.

DuIslingr:
@mstg This is only meant to be temporary until WSL resolves the Find issue with Perl. I fully expect this to be removed at a later date.

DHowett:
To keep with the existing code style and convention, I propose:

```perl
if($^O eq "linux") {
	my $fh;
	open($fh, "/proc/version");
	$File::Find::dont_use_nlink = grep { /Microsoft/ } <$fh>;
	close($fh);
}
```

kirb:
Has this been resolved with recent WSL releases? (Should we require using the latest Insider build, or support stable Windows releases as well?)

DuIslingr:
Insider builds get the wsl updates. You wont see them until the next stable build. As far as this issue is concerned I have not checked recently. Would have to upgrade to latest insider to see but based on WSL changelogs I don't recall seeing a fix for this perl issue. When I have time I will look into it and the other issues I encountered.

kirb:
Sorry, worded that badly. I meant, should we aim to be compatible with 10586 and up, or require the latest insider build? At least, until the Creators Update releases. I suppose everything else seems to work fine anyway on 10586 aside from this, but I haven’t personally used it and we can’t officially vouch for it [here](https://git.io/theosinstall) till we know what we’re guaranteeing compatibility with.

DuIslingr:
Well idk if this has been resolved but if you build s project thats not in the wsl environment it fails to sign it. This is an ldid specific bug though. I worked around this by forcing the build to occur in the wsl environment but this isnt optimal or at least my lame implementation. Latest insider builds allow you to call windows executables. Could try using the windows ldid instead. Ill test this at some point. As far as official support goes i guess that up to you. On one hand if you are on stable you wont get wsl fixes till the next yearly update I guess. While insider does get fixes on a semi regular basis.

DuIslingr:
Basically other than this issue and the $fullname  which we fixed here (https://github.com/DuIslingr/theos/commit/8d642e07ba338c3ec5509b546e3764efd93d7058 )issue theos itself should work as is regardless of insider or stable

kirb:
Rebased + merged both fixes in 4f2856d9e5041bf82e016a30a3e957183bddee80 and a0f6ca125916206c4580b38756e24614015050e3. Thanks!

PR #215, opened 2016-12-25
Add vendor include path to SWIFTFLAGS

kabiroberai:
This fixes a bug due to which Swift cannot find headers in `vendor/include`. Of particular interest is `vendor/include/_Prefix/`, without which Swift doesn't compile in Theos.

No.

This PR doesn't seek to add complete Swift support, and it seems that the Swift deployment target isn't being set correctly. Theos produces the following warning during the linking stage, unless the iOS Deployment Version is set to `latest`:

    ld: warning: object file (/path/to/project/.theos/obj/debug/<arch>/File.o) was built for newer iOS version (10.2) than being linked (7.0)

Without this change, the following error is produced when compiling a Swift file:

    /path/to/theos/Prefix.pch:6:10: error: '_Prefix/NullabilityCompat.h' file not found
            #import <_Prefix/NullabilityCompat.h>
                    ^

Any other comments?
-------------------
Again, this just fixes the bug mentioned above, and doesn't attempt to add the Swift dylibs to `@rpath`. However, if the dylibs are present, the product doesn't crash.

Where has this been tested?
---------------------------
**Operating System:** macOS Sierra 10.12.2 (16C67)

**Platform:** macOS

**Target Platform:** iOS

**Toolchain Version:** Latest

**SDK Version:** iOS 10


kirb:
Merged in 99089ee54084a75400d83077b12087ab624028b3. Thanks!

You missed out the fallback include dir, but I've fixed that in dcb836b09a00ea0dbd371213bcd1fb7a508eb00f.

kabiroberai:
It seems that you accidentally removed `THEOS_VENDOR_INCLUDE_PATH` again in dcb836b, so the issue when compiling still occurs. Also, `THEOS_FALLBACK_INCLUDE_PATH` isn't there either.

Issue #211, opened 2016-11-12
Add lzma support to dm.pl and use it on macOS/Linux


kirb:
dpkg 1.18.11 finally [pulled the plug](https://github.com/guillemj/dpkg/commit/ef6187ded62bb3973e4e57779edc87c3e2904af5) on our saving grace; using `-Zbzip2` or `-Zlzma` is now an error.

The most straightforward way to “fix” this would be to just replace it with [dm.pl](https://github.com/theos/theos/blob/master/bin/dm.pl) and add lzma support to it.

(Telesphoreo’s dpkg intentionally does not support xz due to [various flaws](http://www.nongnu.org/lzip/xz_inadequate.html) with xz and LZMA2.)

## Error output
```
dpkg-deb: error: obsolete compression type 'lzma'; use xz instead

Type dpkg-deb --help for help about manipulating *.deb files;
Type dpkg --help for help about installing and deinstalling packages.
make: *** [internal-package] Error 2
```

## Workaround
For the moment, you’ll need to just use 1.18.10.

For Homebrew, you can downgrade like so:

```shellsession
$ brew remove dpkg  # remove latest dpkg
$ brew install --force-bottle https://raw.githubusercontent.com/Homebrew/homebrew-core/7a4dabfc1a2acd9f01a1670fde4f0094c4fb6ffa/Formula/dpkg.rb  # install dpkg as a bottle from the old commit
$ brew pin dpkg  # block homebrew from updating dpkg till you `brew unpin dpkg`
```

(Updated 01/03/17 using @Fr0stDev’s workaround because Debian delete old source releases. 1.18.10 is from July 2016.)

Fortunately, seems Debian/Ubuntu are safe for the moment: https://launchpad.net/dpkg/+packages. But I would expect them to be updating sometime soon.

----

Partly dupes #197.

DHowett:
To facilitate this, I've transferred **DHowett/dm.pl** to **theos/dm.pl**.

kirb:
Cheers :+1:

Hax0rDoge:
Following these instructions let me to an poor where there was an undefined variable "@@HOMEBREW_CELLAR@@". [Here](https://stackoverflow.com/questions/7907055/bash-usr-local-bin-heroku-usr-local-bin-ruby-bad-interpreter-no-such-file) is a solution and it worked for me. Just leaving this here in case anyone else also gets this error.

kirb:
Whoops. Updated to include a sed step.

[...]

PR #217, opened 2016-12-29
[Makefile] THEOS is now using GZIP on macOS Sierra

ijapija00:
Using lzma on macOS Sierra, will throw 'subprocess dpkg-deb --fsys-tarfile returned error exit status 2', but using xz is impossible since Cydia's dpkg doesn't support that. Instead, with the latest dpkg update, using 'dpkg-deb -bZgzip ...' solves the lzma problem on macOS Sierra and at the same time allows installations on iOS Devices.


kirb:
Thanks for taking the time to make a PR but please understand that there is an existing discussion about this (#211) and forcing gzip is not a proper solution.

PR #225, opened 2017-01-28
Support true parallel builds

kirb:
Enables true parallel builds, either when specifying the `-j` flag, and by default using as many logical cores (e.g. 8 on a Core i7, 4 on a Core i5, 2 on an iPhone 6s). ~~We just need to pass `$(MFLAGS)` to sub-make calls, so that the job server flag is included (so sub-makes can communicate with the parent and each other, ensuring they only use the specified number of threads at a time).~~ Without the flag, sub-makes revert to working serially:

```
make[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```

No

Any other comments?
-------------------
* ~~Usually make is meant to inject the jobserver flags when it sees `$(MAKE)` being used, but that wasn’t working for me, even when adding a `+` at the start of the line as the warning suggests. However, there is `$(MFLAGS)`, which is ”for historical compatibility,” but does exactly what we want.~~ I was Doing It Wrong™. `$(MAKEFLAGS)` is a set of flags implicitly passed down to sub-makes via the environment; it doesn’t have to be passed in manually.

* This also adds `--output-sync=target` to the MAKEFLAGS, so the output doesn't look like vomit if there are errors (although it sure makes it look less fast). This isn't supported by (you guessed it) Apple's ancient version of make. ~~So for now, this requires macOS users to:~~ As such parallelism isn’t supported on macOS unless make is updated to the latest via:

  ```shellsession
  $ brew install make
  $ alias make=gmake
  ```

* ~~At the moment there are still some cases where it might output:~~ Fixed.

  ```
  make[4]: warning: -jN forced in submake: disabling jobserver mode.
  ```

* ~~I want to make parallelism completely implicit if at all possible. As long as the user hasn’t passed a `-j` flag of their own on the command line, Theos should automatically determine the appropriate number of jobs to run at once, and then re-run make with the same command line plus, eg, `-j 4` for a 4-thread CPU.~~ Done.

Where has this been tested?
---------------------------
* **Operating System:** macOS 10.12.3
* **Target Platform:** iOS
* **Toolchain Version:** Xcode 8.1
* **SDK Version:** iOS 10.1


kirb:
I’m confident in this now. I haven’t been able to get it to complain about passing `-j` at the same time as `--jobserver-auth` (I check for that flag in `$(MAKEFLAGS)`, which seems to be what the make manual recommends), and it now determines a reasonable default number of threads (equal to number of logical cores on your CPU) to pass to second-level `make` invocations. It also checks for Apple’s super old `make` and disables parallelism, and can be disabled with `THEOS_USE_PARALLEL_BUILDING = no`. I guess it can be merged 😛

PR #246, opened 2017-04-28
[theos/bin/logos.pl] Added unicode support* for theos.

BlueCocoa:
[theos/bin/logos.pl] Added unicode support* for theos. 

Added unicode support* for theos.

*, You can use any allowed unicode characters in Objective-C.

In real-life, Objective-C accepts unicode characters like Chinese, Japanese, Arabic,  so on and so forth in both class name and method name. 

However, the regex in ```theos/bin/logos.pl``` only matches ```[\$_\w]```, which would emit error if there is any unicode character.

The error message goes like,

```$ make
> Making all for tweak UnicodeHook…
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (armv7)…
Tweak.xm:3:1: error: missing context for method declaration
- (id)中文方法名 {
^
1 error generated.
```

So, I just modified the regex pattern, patterns like ```[\$_\w]``` were changed to ```[^\s]```. IMHO, this is good, after all, I believe probably no one would write meaningless ASCII characters in these commands. Even if they do that, the file won't compile.

Affected commands are listed below.

- %hook
- %subclass
- %group
- %class
- %c

Applied these modifications, we got

```
$ export THEOS=~/Git/theos
$ make clean
==> Cleaning…
$ cat Tweak.xm 
%hook UnicodeHook可行

- (id)中文方法名 {
    return %orig;
}

- (id)アルパカ {
    return %orig;
}

- (id)بالعربية {
    return %orig;
}

%end
$ make
> Making all for tweak UnicodeHook…
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (armv7)…
==> Linking tweak UnicodeHook (armv7)…
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of iOS 7 [-Wdeprecated]
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (arm64)…
==> Linking tweak UnicodeHook (arm64)…
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of iOS 7 [-Wdeprecated]
==> Merging tweak UnicodeHook…
==> Signing UnicodeHook…
$ 
```

AFAIK, no such related open issues.

Before these changes were made, Logos emits error massages.

Tweak.xm,
```
%hook UnicodeHook

- (id)中文方法名 {
}

%end

```

```$ make
> Making all for tweak UnicodeHook…
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (armv7)…
Tweak.xm:3:1: error: missing context for method declaration
- (id)中文方法名 {
^
1 error generated.
```

Any other comments?
-------------------
I guess no. Maybe you can refer to my blog post on this topic, [Theos Hook Objective-C classes / methods with Unicode Characters](https://blog.0xbbc.com/2017/04/theos-hook-objective-c-classes-methods-with-unicode-characters/). 

The last screenshot is how the compiled dylib looks like in Hopper, which suggests these modifications
indeed work. 

Where has this been tested?
---------------------------
**Operating System:**
macOS 10.12.4

**Platform:** …
macOS

**Target Platform:** …
iOS/iPhone

- armv7
- arm64

**Toolchain Version:**
toolchain comes with Xcode 8.3 (8E162)

**SDK Version:**
```
$ xcrun --sdk iphoneos --show-sdk-path
/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS10.3.sdk
```

### DIFF
--- a/bin/logos.pl
+++ b/bin/logos.pl
@@ -224,7 +224,7 @@ foreach my $line (@lines) {
 		$directiveDepth = 0 if(!@directiveDepthTokens);
 		$directiveDepth = $depthsForCurrentLine{join(':', @directiveDepthTokens)} if @directiveDepthTokens;
 
-		if($line =~ /\G%hook\s+([\$_\w]+)/gc) {
+		if($line =~ /\G%hook\s+([^\s]+)/gc) {
 			# "%hook <identifier>"
 			fileError($lineno, "%hook does not make sense inside a block") if($directiveDepth >= 1);
 			nestingMustNotContain($lineno, "%hook", \@nestingstack, "hook", "subclass");
@@ -236,7 +236,7 @@ foreach my $line (@lines) {
 			$currentClass = $currentGroup->addClassNamed($1);
 			$classes{$currentClass->name}++;
 			patchHere(undef);
-		} elsif($line =~ /\G%subclass\s+([\$_\w]+)\s*:\s*([\$_\w]+)\s*(\<\s*(.*?)\s*\>)?/gc) {
+		} elsif($line =~ /\G%subclass\s+([^\s]+)\s*:\s*([^\s]+)\s*(\<\s*(.*?)\s*\>)?/gc) {
 			# %subclass <identifier> : <identifier> \<<protocols ...>\>
 			fileError($lineno, "%subclass does not make sense inside a block") if($directiveDepth >= 1);
 			nestingMustNotContain($lineno, "%subclass", \@nestingstack, "hook", "subclass");
@@ -265,7 +265,7 @@ foreach my $line (@lines) {
 			$classes{$classname}++;
 
 			patchHere(undef);
-		} elsif($line =~ /\G%group\s+([\$_\w]+)/gc) {
+		} elsif($line =~ /\G%group\s+([^\s]+)/gc) {
 			# %group <identifier>
 			fileError($lineno, "%group does not make sense inside a block") if($directiveDepth >= 1);
 			nestingMustNotContain($lineno, "%group", \@nestingstack, "group");
@@ -284,7 +284,7 @@ foreach my $line (@lines) {
 				$patchSource = Patch::Source::Generator->new($capturedGroup, 'declarations');
 			}
 			patchHere($patchSource);
-		} elsif($line =~ /\G%class\s+([+-])?([\$_\w]+)/gc) {
+		} elsif($line =~ /\G%class\s+([+-])?([^\s]+)/gc) {
 			# %class [+-]<identifier>
 			@firstDirectivePosition = ($lineno, $-[0]) if !@firstDirectivePosition;
 
@@ -300,7 +300,7 @@ foreach my $line (@lines) {
 			}
 			$classes{$classname}++;
 			patchHere(undef);
-		} elsif($line =~ /\G%c\(\s*([+-])?([\$_\w]+)\s*\)/gc) {
+		} elsif($line =~ /\G%c\(\s*([+-])?([^\s]+)\s*\)/gc) {
 			# %c([+-]<identifier>)
 			@firstDirectivePosition = ($lineno, $-[0]) if !@firstDirectivePosition;
 
@@ -321,7 +321,7 @@ foreach my $line (@lines) {
 			$xtype = $2 if $2;
 			$newMethodTypeEncoding = $xtype;
 			patchHere(undef);
-		} elsif($currentClass && $line =~ /\G([+-])\s*\(\s*(.*?)\s*\)(?=\s*[\$\w:])/gc && $directiveDepth < 1) {
+		} elsif($currentClass && $line =~ /\G([+-])\s*\(\s*(.*?)\s*\)(?=\s*[^\s])/gc && $directiveDepth < 1) {
 			# [+-] (<return>)<[X:]>, but only when we're in a %hook.
 
 			# Gasp! We've been moved to a different group!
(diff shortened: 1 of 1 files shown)


uroboro:
Looks like a good improvement. Any chance of adding support for %property (and %hookf if applicable)?

candoizo:
Hi, I have tested these changes on my machine and hope some feedback would be welcome

Platform: Linux , same targets
9.2 SDK

I was able to successfully compile an example tweak that used unicode in the method / variable names.

On the other hand, I then tried to switch over to some of my recent projects that do not have special characters, which were built with the current pull of theos and it was unable to compile using these logos.pl changes. Switching back to the currently shipped logos made everything work as expected, all the projects only used regular %hooks.

Error logs (similar in most cases where this happens) : https://ghostbin.com/paste/9acp2

Heres the source from an .xm of a basic tweak that won't compile using these changes, maybe it would help spot the issue

```
%hook SBDashBoardPasscodeViewController
-(void)passcodeLockViewEmergencyCallButtonPressed:(id)arg1 { //the line that spawns errors
       %orig;
  UIAlertController * alert=   [%c(UIAlertController) alertControllerWithTitle:@"Power Options"
					    message:nil
                                            preferredStyle:UIAlertControllerStyleActionSheet];
   UIAlertAction* cancel = [%c(UIAlertAction) actionWithTitle:@"Cancel"
              				 style:UIAlertActionStyleCancel
              	                          handler:^(UIAlertAction * action)
                                                 {
                                              [alert dismissViewControllerAnimated:YES completion:nil];
                                                }];
         [alert addAction:cancel];
	[self presentViewController:alert animated:YES completion:nil];

}

%end
```

kirb:
Interesting, so it's matching far more than it should be. Thoughts @uroboro? I need to re-read the logic that matches method names but presumably we can make it match up to the first colon seen? This kinda makes me unsure about other corner cases of now-broken syntax we might not be aware of.

DHowett:
So!
Method names are matched in parts by successive regexes.

`- (type)name` is matched all at once, and only if that's found do we look for `:(type)arg`, `name`, `:(type)arg`, `...`

bin/logos.pl
```diff
@@ -236,7 +236,7 @@ package main;
 			$currentClass = $currentGroup->addClassNamed($1);
 			$classes{$currentClass->name}++;
 			patchHere(undef);
-		} elsif($line =~ /\G%subclass\s+([\$_\w]+)\s*:\s*([\$_\w]+)\s*(\<\s*(.*?)\s*\>)?/gc) {
+		} elsif($line =~ /\G%subclass\s+([^\s]+)\s*:\s*([^\s]+)\s*(\<\s*(.*?)\s*\>)?/gc) {
```
DHowett: The problem is likely here: matching "not-space" followed by a `:` will match `x:y:z:a:b:c` because it's greedy.

You could experiment with replacing the `+` with `+?` in each of the changed sections.

DHowett:
I think the ideal would be to use the character classes containing all languages' standard "alpha"-equivalent characters. Using `[^\s]` is likely to cause no end of headache.

DHowett:
Finally, and sorry for the spam, you probably want `\p{Alnum}` combined with `$` and `_`.

```
   \p{Alnum}               \p{XPosixAlnum} (118_820)
   \p{Alpha}               \p{XPosixAlpha} (= \p{Alphabetic=Y})
                             (118_240)
```

[perluniprops](https://perldoc.perl.org/perluniprops.html#Properties-accessible-through-%5cp%7b%7d-and-%5cP%7b%7d)

[perlrecharclass](https://perldoc.perl.org/perlrecharclass.html)

BlueCocoa:
Sorry about not noticing the emails, I just applied some changes to it and it now works on my computer with these codes below. And I've written the detail (in simple words, `\p{Alnum}` and its variants didn't work as expected) in the commit, but there may be some other cases that cannot compile. 

```
BlueCocoa:char bluecocoa$ make
> Making all for tweak char…
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (armv7)…
==> Linking tweak char (armv7)…
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of iOS 7 [-Wdeprecated]
==> Preprocessing Tweak.xm…
==> Compiling Tweak.xm (arm64)…
==> Linking tweak char (arm64)…
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of iOS 7 [-Wdeprecated]
==> Merging tweak char…
==> Signing char…
```

```logos
%hook UnicodeHook可行

- (id)中文方法名:(id)arg1 参数:(id)arg2 {
	return %orig;
}

- (id)中文方法名:(id)arg1 
           参数2:(id)arg2 {
	return %orig;
}

- (id)アルパカ {
	return %orig;
}

- (id)بالعربية {
	return %orig;
}

- (void)passcodeLockViewEmergencyCallButtonPressed:(id)arg1 {
	%orig;
	UIAlertController * alert = [%c(UIAlertController) alertControllerWithTitle:@"Options"
																		message:nil
																 preferredStyle:UIAlertControllerStyleActionSheet];
	UIAlertAction * cancel = [%c(UIAlertAction) actionWithTitle:@"Cancel"
											 		      style:UIAlertActionStyleCancel
								                        handler:^(UIAlertAction * action) {
									[alert dismissViewControllerAnimated:YES completion:nil];
	}];
	[alert addAction:cancel];
	[self presentViewController:alert animated:YES completion:nil];
}

%end
```

BlueCocoa:
Now it got the same build log for all 3 tests on Travis CI as some merged pull requests, e.g #253. Presumably this patch now reaches its purpose.

PoomSmart (approved):

PR #258, opened 2017-08-08
WSL Support

DuIslingr:
Fixes various WSL build problems. LDID will fail upon merging because of an inherent problem with Windows and files being open in processes. This moves the build process inside WSL to avoid this problem. Also we cannot apply linux file permissions to files outside of WSL which this also resolves an issue when packaging files. As of right now normal fakeroot doesn't work. fakeroot-tcp needs to be installed in place of fakeroot on WSL to avoid another issue when packaging. I believe that is everything of note at this time.

Where has this been tested?
---------------------------
**Operating System:** Windows 10 build 16257 (WSL running Arch Linux instead of Ubuntu)

**Platform:** WSL

**Target Platform:** iOS 10

**Toolchain Version:** Clang 4.0.1 with latest CCTools Port https://github.com/tpoechtrager/cctools-port
As it supports latest tbd version it resolves the problem with coolstars toolchain not supporting SDK 10.
**SDK Version:** iPhoneOS10.1.sdk provided in theos/sdks


kirb:
Merged via rebase with the bunch of other merges happening in #291. Thanks :+1:

Issue #176, opened 2016-07-30
Add a gitignore when creating projects


kirb:
Reduce the possibility of newcomers unknowingly committing build artefacts by writing a gitignore while creating a new (non-child) project.

```
.theos/
packages/
```

PR #279, opened 2018-01-23
[nic.pl] Add git initialization support

CPDigitalDarkroom:
Allow users to add init_git to their ~/.nirc configuration to have a git repository initialized during setup.

This implements initializing a git repository for the created project if the user has enabled 'init_git' in ~/.nirc 

Not that I've seen
…

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Operating System:** macOS 10.13.2

**Toolchain Version:** … Xcode 9

**SDK Version:** … 10.3

### DIFF
--- a/bin/nic.pl
+++ b/bin/nic.pl
@@ -157,6 +157,8 @@ my $dirname = lc($clean_project_name);
 $NIC->build($dirname);
 chdir($cwd);
 
+my $isSubproject = 0;
+
 my @makefiles = ("GNUmakefile", "makefile", "Makefile");
 my $makefile;
 map { $makefile = $_ if -e $_; } @makefiles;
@@ -168,6 +170,7 @@ if($makefile) {
 		my $alreadyHas = 0;
 		map {$alreadyHas++ if /^\s*SUBPROJECTS.*$dirname/;} @lines;
 		if($alreadyHas == 0) {
+			$isSubproject = 1;
 			print "Adding '$project_name' as an aggregate subproject in Theos makefile '$makefile'.",$/;
 			my $newline = "SUBPROJECTS += $dirname";
 			my $i = 0;
@@ -183,6 +186,17 @@ if($makefile) {
 	}
 	untie(@lines);
 }
+
+my $initializeGit = $CONFIG{'init_git'};
+$initializeGit = 0 if !$initializeGit;
+
+if($initializeGit == 1 &! $isSubproject) {
+	chdir($dirname);
+	my $init = `git init; git add -A; git commit -m "Initial Commit"`;
+} elsif($initializeGit == 1 && $isSubproject) {
+	my $init = `git add '$dirname'; git commit -m "Instantiated Subproject $clean_project_name"`;
+}
+
 print "Done.",$/;
 
 sub promptIfMissing {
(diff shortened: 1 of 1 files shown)


bin/nic.pl
```diff
@@ -183,6 +186,17 @@
 	}
 	untie(@lines);
 }
+
+my $initializeGit = $CONFIG{'init_git'};
+$initializeGit = 0 if !$initializeGit;
+
+if($initializeGit == 1 &! $isSubproject) {
```
uroboro: What's this `&!` operator?

CPDigitalDarkroom: Oops, right, that's not a perl operator. It worked correctly during my testing so I didn't even think twice.

kirb:
Good idea. Sorta ties in with #176 which I haven't gotten around to yet. Wondering if we should add this as a prompt ("Initialise a Git repository for the project? [yes]").

CPDigitalDarkroom:
Prompting the user is a great idea. I've updated the pull request with it as well changing the config key name to 'request_git'. That way it won't prompt if a user has already explicitly marked wanting git.

bin/nic.pl
```diff
@@ -183,6 +187,23 @@
...
+	$requestsGit = (uc($requestsGit) eq "N") ? 0 : 1;
+
+	if($requestsGit) {
+		chdir($dirname);
+		my $init = `git init; git add -A; git commit -m "Initial Commit"`;
+	}
+} else {
+	# Probably don't want to prompt for git in subproject. Only if user has already requested
+	# git in the config do we add this commit message otherwise we don't know if git is setup
+	if($requestsGit) {
```
kabiroberai: Perhaps you could also run `git add` if the main project already has a git repository.

andrewwiik:
maybe also add the option to include a Travis-CI config for the package?

andrewwiik:
Also maybe add the option to make it automagically make a commit when you run make package FINALPACKAGE=1 with version number in the commit message so you automagically have commits for each public release to roll back on if needed?

bin/nic.pl
kabiroberai: i.e. if `$requestsGit` is true or `.git/` exists

uroboro:
Making a commit when packaging is a bit out of scope of the initial proposal. This affects the creation of a project, making changes only to `nic.pl`.

kabiroberai:
@andrewwiik imo the FINALPACKAGE detection idea may be a bit intrusive, because if you were to repeat the same make command multiple times it would unintentionally result in several commits.

kirb:
@andrewwiik, (I guess you deleted your comments…) I'd rather Theos not be super intrusive/opinionated on things like whether you should have CI, which CI provider to use, etc. (Also we're not really well-suited for CI at this point, and Travis's macOS VMs are quite slow to start as compared to their Linux ones  Nice to have but I removed it from TypeStatus after a while because it was just another worry for a solo-maintainer project.) Assuming you want a Git repo is still a bit intrusive/opinionated but at least it's common and useful enough that we might as well offer it to save you a few seconds. I'm thinking NIC could do with a `base.nic.tar` that gets processed first so you can add your own personal boilerplate stuff like .travis.yml to it.

Issue #243, opened 2017-03-22
CGFloat %property wrongly assumes float, crashes on 32-bit


CoBrA2168:
What are the steps to reproduce this issue?
-------------------------------------------
1. Create a %property (with nonatomic, assign) into a hooked class that is of type CGFloat
2. Attempt to access this property from within the hooked code (i.e. self.MyFloat) on a 32-bit device.

What happens?
-------------
The dylib will crash.

What were you expecting to happen?
----------------------------------
The dylib accesses the float value as expected.

Any logs, error output, etc?
----------------------------
You might see this in the logs:
`[NSISRestrictedToZeroMarkerVariable floatValue]: unrecognized selector sent to instance 0x197a33e0`

Any other comments?
-------------------
@uroboro believes that this is due to the way %property is generated for CGFloat values and that is wrongly assumed to be of type float.  See: https://github.com/theos/theos/blob/master/bin/lib/Logos/Generator/Base/Property.pm for the definition of a %property that is a CGFloat.  My code will run perfectly fine on 64-bit devices (iPhone 6s on 9.3.3, iPhone 6s on 10.2, iPad Air on 9.3.2) but fails on 32-bit devices (iPad 4 on 9.3.3).

What versions of software are you using?
----------------------------------------
**Operating System:** 9.3.3 (iPad3,4 with Home Depot jailbreak)

**Toolchain Version:** 9.3

**SDK Version:** 9.3

kirb:
We discussed this when it was filed – `+[NSValue valueWithBytes:objCType:]` and `-[NSValue getValue:]` were brought up as a way we can store primitive values generically. As we know the exact type, we can get back exactly what was stored rather than approximating it with NSNumber like we do now.

PR #282, opened 2018-01-27
[Logos] %property rewrite.

uroboro:
Rewrites %property so its attributes are parsed only once in logos.pl and the "backend" uses the sigil name generator as well as aligning its methods to the rest of the directives (`getters_setters` -> `definitions`). The code generator wraps the objects into NSValues instead of supporting a subset of types with specific APIs from NSNumber and NSValue.

#243

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** macOS Sierra

**Platform:** N/A

**Target Platform:** N/A

**Toolchain Version:** N/A

**SDK Version:** N/A

### DIFF
--- a/bin/lib/Logos/Generator/Base/Property.pm
+++ b/bin/lib/Logos/Generator/Base/Property.pm
@@ -2,222 +2,37 @@ package Logos::Generator::Base::Property;
 use strict;
 use Logos::Util;
 
-sub key {
+sub getterName {
 	my $self = shift;
 	my $property = shift;
-
-	my $build = "_logos_property_key\$" . $property->group . "\$" . $property->class . "\$" . $property->name;
-
-	return $build;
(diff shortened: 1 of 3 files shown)


uroboro:
Not longer applicable, moving code to Logos repo.

kirb:
For reference what we discussed:

* For `assign` properties, we box the value into an NSValue with policy `nonatomic, retain`.
* Otherwise, we assume it's an ObjC type and directly assign it as the value. 
* Also, `strong` will be an alias of `retain`; `unsafe_unretained` to `assign`; `weak` isn't possible and will be an error. 

Will be reopened as a PR on [theos/logos](https://github.com/theos/logos)

uroboro:
The last point is already implemented in the existing changes.

PR #302, opened 2018-03-18
Compile on-device with Electra

sbingner:

This changes (hopefully) all external calls to not rely on #! - primarily perl

Not that I saw

Any other comments?
-------------------
It shows a number of commits in the past from my branch but they are not included in this pull request.


makefiles/package.mk
```diff
@@ -56,7 +56,7 @@ ifeq ($(_THEOS_FINAL_PACKAGE),$(_THEOS_TRUE))
 VERSION.EXTRAVERSION = $(if $(PACKAGE_BUILDNAME),+$(PACKAGE_BUILDNAME))
 _THEOS_PACKAGE_DEFAULT_VERSION_FORMAT = $(THEOS_PACKAGE_BASE_VERSION)$(VERSION.EXTRAVERSION)
 else
-VERSION.INC_BUILD_NUMBER = $(shell THEOS_PROJECT_DIR="$(THEOS_PROJECT_DIR)" "$(THEOS_BIN_PATH)/package_version.sh" -N "$(THEOS_PACKAGE_NAME)" -V "$(__BASEVER_FOR_BUILDNUM)")
+VERSION.INC_BUILD_NUMBER = $(shell bash THEOS_PROJECT_DIR="$(THEOS_PROJECT_DIR)" "$(THEOS_BIN_PATH)/package_version.sh" -N "$(THEOS_PACKAGE_NAME)" -V "$(__BASEVER_FOR_BUILDNUM)")
 _THEOS_PACKAGE_DEFAULT_VERSION_FORMAT = $(THEOS_PACKAGE_BASE_VERSION)-$(VERSION.INC_BUILD_NUMBER)$(VERSION.EXTRAVERSION)
```
DHowett: I'm somewhat surprised that we ended up shipping "`FINALPACKAGE`". The whole point of the composable versioning system was to make that second part not required ;P

DHowett:
why is the shebang not correct here? does electra not support shebangs? how do _any_ package maintscripts work?

Is it reasonable to introduce changes to Theos to support a jailbreak that doesn't support _anything_?

makefiles/package.mk
DHowett: it also doesn't seem safe to assume `bash` everywhere. Especially if the script isn't using any bashisms?

sbingner:
That's right no shebangs work... I figured it was probably ok to make it not use them since there should not be a place where it would break anything.  dpkg apparently has something where it reads the file before execing to see if it is a shebang, and if so executes it using the interpreter specified

makefiles/package.mk
sbingner: The shebang in the script is #!/usr/bin/bash which is why I used bash

kirb:
That workaround really should be implemented in bash as well, I asked coolstar to do it a while back. Was hoping to not need to do this in Theos but we're still waiting, so I'm planning to merge this so Theos works-ish for now and we can revert later on.

makefiles/package.mk
kirb: I've been hoping to revisit the FINALPACKAGE implementation, it hasn't changed since I merged rpetrich's branch a few years ago. I figured it was convenient shorthand for "I'm releasing this", and it also calls `plutil -convert binary1` and `pincrush` (that too needs revisiting…). 

We force SHELL = bash at the top of common.mk, I don't know of any realistic Unix-alike system that doesn't ship bash, and I haven't seen anyone try to run Theos on anything more interesting than macOS/iOS or Linux/Linux-alike. So it's safe to assume bash as we already have a hard dependency on it anyway.

sbingner: It should probably be $(SHELL) instead of "bash" then - I missed that variable.

sbingner:
With bash patched this isn't necessary... I made a different pull request #303 with the only changes needed

PR #303, opened 2018-03-20
Use dm.pl when possible on iOS

sbingner:
Changes Darwin-arm to use dm.pl when it is present and perl supports LZMA

no

Any other comments?
-------------------
With the patched bash this should be enough to make theos work on Electra

Where has this been tested?
---------------------------
**Operating System:** iOS, macOS

**Target Platform:** iphone

### DIFF
--- a/makefiles/platform/Darwin-arm.mk
+++ b/makefiles/platform/Darwin-arm.mk
@@ -3,7 +3,12 @@ _THEOS_PLATFORM_LOADED := 1
 THEOS_PLATFORM_NAME := iphone
 
 _THEOS_PLATFORM_DEFAULT_TARGET := iphone
-_THEOS_PLATFORM_DPKG_DEB := dpkg-deb
+_THEOS_PLATFORM_DPKG_DEB := $(shell \
+	PATH=$$PATH:$(THEOS_BIN_PATH); \
+	DM=$$(which dm.pl); \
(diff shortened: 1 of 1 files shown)


makefiles/platform/Darwin-arm.mk
```diff
@@ -3,7 +3,13 @@ _THEOS_PLATFORM_LOADED := 1
 THEOS_PLATFORM_NAME := iphone
 
 _THEOS_PLATFORM_DEFAULT_TARGET := iphone
-_THEOS_PLATFORM_DPKG_DEB := dpkg-deb
+_THEOS_PLATFORM_DPKG_DEB := $(shell \
+	PATH=$$PATH:$(THEOS_BIN_PATH); \
+	DM=$$(which dm.pl); \
+	[ -z "$${DM}" -a -f "$(THEOS_BIN_PATH)/dm.pl" ] && DM="$(THEOS_BIN_PATH)/dm.pl"; \
+	[ ! -z "$${DM}" ] && ($(PERL) -MIO::Compress::Lzma -e '1' 2>/dev/null) \
```
kabiroberai: Does `$(PERL)` exist without #302?

kirb: I replaced this with just `perl`, I think it’s just a remnant of the workarounds. Should work without it now that we have the hashbang workaround in bash.

PR #367, opened 2019-01-04
Add support for building archives

### LINKED ISSUE #310: Feature request: build static library
would be useful when developing a big project.

sbingner:
This allows you to create an archive (.a) AKA static library

closes #310


DHowett:
I believe I had outstanding work on this that enhanced the “library”
instance type. I think it makes more sense that way; perhaps via
xxx_LINKAGE_TYPE=[static|dynamic]?

DHowett:
Fair point about the name. Perhaps `xxx_LIBRARY_TYPE=[static|dynamic|both]` with a default of `dynamic`? I do think that since they're really two faces of the same type of output, they belong on the same instance type. I don't think "archive" is a meaningful concept at this layer, especially since that precludes "superobject" libraries -- a single `.o` file -- which might be a possible implementation for `LIBRARY_TYPE=static`?

There's a few interesting avenues to explore after that -- `xxx_yyy_LINKAGE_TYPE=static` could dictate how `xxx` consumes `yyy`?

sbingner:
OK, this does what you said... what do you think?  Not totally happy with a couple things like the echo output but it seems good to me.  I could add logic to not include the archive rules when they won't be used, but wasn't sure if it was worth it

DHowett:
I like how this looks. I imagine the output is somewhat like:

```
Linking for library x ...
Linking static for library x ...
```

Having a global defaulted `THEOS_LINKAGE_TYPE` is interesting, actually. I didn't think about what it'd mean to have a project with multiple libraries in _all_ switching between static/both/dynamic, but I think it makes sense.

`..ARCHIVE_EXTENSION` is a platform-target concern.

Overall, though, I like it.

DHowett:
(For `ARCHIVE_EXTENSION`, consider that on Windows it's `.lib` -- and Windows always has to be the odd one out. There's a few other concerns there, too, as `both`-type linkage doesn't work right there. Producing a DLL _also_ produces a LIB, but the LIB is an import library.)

sbingner:
I don’t think I can fix it for windows, I don’t have a build environment set up for windows anywhere :|

kirb:
Merged into [next](https://github.com/theos/theos/compare/master...next) branch. Needs testing.

kabiroberai:
@kirb should we close this now since the next branch has been merged into master?

kabiroberai:
Related: a bit late to review this now, but we should generalise this to frameworks as well since it's legal for them to be static too (and that's one potential way to fix the Orion bootstrapping issue)

kirb:
Yep this is meant to be closed now. @kabiroberai, can you make an issue for that?

PR #371, opened 2019-01-19
Change default C version

coolstar:
C99 is very old and doesn't support a lot of newer features. Set the default to GNU11

Set the default C version to GNU11 rather than C99. Tested with inline assembly (asm() function) to make sure compilation succeeds (unlike setting it in CFLAGS since setting it there didn't work).

Where has this been tested?
---------------------------
**Operating System:** macOS 10.14

**Platform:** Darwin

**Target Platform:** iOS

**Toolchain Version:** XCode 10.1

**SDK Version:** 12.1


kirb:
I read the clang docs a few weeks ago and realised gnu14 is actually the default if no `-std=` flag is passed at all. I think it’d be a good idea to just remove `-std=c99` entirely and trust the compiler’s default, which no doubt will be updated as the implementation of newer features becomes more stable. I don’t think Xcode even passes `-std=` by default unless explicitly overridden in the project config, so we end up matching this behavior effectively.

kirb:
Implemented that change with 53ca676.

Issue #286, opened 2018-02-07
Theos for the App Store


kirb:
Theos should be able to build apps for releasing on the App Store. This would allow apps to be built on a non-Mac platform. Theos can already support building a full-blown real-deal app, including app extensions and an Apple Watch extension, without using anything that’d be considered private API. It should be completely possible to install a built app with a dev certificate on a connected phone, to debugging it with lldb (maybe even over wifi?), to uploading it to iTunes Connect to be pushed to TestFlight or the App Store. I don’t think there’s any reason Apple can reject apps built and submitted using Theos, if it all works properly.

### Prelude stuff:
* **Merge #278 so Swift works again.**
* **Release a working Swift compiler for building for Darwin/iOS on Linux, based on the official Swift.org builds for Linux.**  
   Probably also Cygwin but we’re discouraging use of Cygwin now.
* **Complete the unifying of the Darwin targets (#242).**  
   This would make it far less painful to create some App Store targets based on top of the existing `iphone`, `watchos`, etc targets and keep every platform and target in sync.
* **Build iOS/tvOS app and app extension templates.**  
   We already have application, and application_swift is coming in theos/templates#7. Also create tvos_application and tvos_application_swift, as well as templates for each kind of app extension that Theos can currently build without trouble (only in either ObjC or Swift, not both, to avoid clogging up the NIC list).

### Then, on a development branch:
* **Build a target based on the current `iphone` targets that set extra defaults for App Store distribution.**  
   Such as the app being staged to `/Payload` rather than `/Applications`, to enable Bitcode generation for release builds (required to submit a watchOS/tvOS app; not required but recommended for iOS apps), and possibly to disable codesigning on Theos’s part (see below).
* **Implement or find an existing cross-platform tool to generate .car asset catalogs.**  
   Seems like iOS 11 requires the app icon to be in a .car now? Either way it’d be cleaner to support this, both for App Store apps and any bundle built for a jailbreak package.
* **Figure out how we’ll support storyboards/xibs.**  
   While there’s some experimental storyboardc/nib compilers out there, an Interface Builder-like app would still be needed to create them. In a discussion we had a few weeks ago, we figured a primitive storyboard compiler would be fine to initially provide only Good Enough support for launch screens. Everything else can be achieved in code until a cross-platform interface builder app exists and is mature enough.
* **Add `ipa` packaging target.**  
   This would just zip up the staging directory and rename the file to end in .ipa.
* **Add `ipa` install target.**
[...]

PR #415, opened 2019-06-01
[package] Add ipa packaging support

uroboro:
Adds ipa packaging for #286.

### DIFF
--- a/makefiles/install/ipa_local.mk
+++ b/makefiles/install/ipa_local.mk
@@ -1,5 +1,0 @@
-internal-install:: internal-install-check
-	$(ECHO_INSTALLING)appinst "$(_THEOS_PACKAGE_LAST_FILENAME)" &> /dev/null$(ECHO_END)
-
-internal-uninstall::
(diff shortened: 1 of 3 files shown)


uroboro:
This implementation relies on a single application being present in the project, but I hope that is the general case for making application projects. Otherwise, this will be delayed even more to implement mechanisms to package multiple apps from a single Theos project.

kabiroberai:
IMO assuming that there’s only one app in the project is fine, since Xcode itself does the same thing when you archive a target. In particular, if it detects that any dependencies have install destinations outside the target’s app dir (/Applications/[app name].app), it won’t let you export the archive as an ipa file.

leptos-null:
This PR looks good, and very helpful. 
The requirement of a `control` file is a dpkg concept, where an `Info.plist` is the somewhat-equivalent for an ipa. The two fields from the `control` file (Package, Version) are also present in a bundle `Info.plist` (CFBundleIdentifier, CFBundleVersion). I understand this is a fairly small detail, and that a method for reading plists is not portably available.

kirb:
Merged into [next](https://github.com/theos/theos/compare/master...next) branch. Needs testing.

Issue #430, opened 2019-08-12
Is it possible to use LDFLAGS to add 3rd party framework


liuxuan30:
So I read https://github.com/theos/theos/issues/110 that how theos adds 3rd party framework.

I have two questions:

1. Can we use LDFLAGS and CFLAGS to add a 3rd party framework?

I tried to use `MyTweak_LDFLAGS += -F./Frameworks/MyFramework.framework`, it however said cannot find the header. If I delete the code and just compile, it seems passes though.

2. I also tried to use `MyTweak_EXTRA_FRAMEWORKS = MyFramework`, copy `MyFramework.framework` into `theos/lib/`

and in tweak.xm, import like this:
```
#import <MyFramework/MyFramework.h>   
```
still throws the same error:
>Tweak.xm:1:9: fatal error: 'MyFramework/MyFramework.h' file not found   

Can someone help? Is this a bug, or my misusage? Thanks

kirb:
`-F` (and `-I`, `-l`) are search paths, they allow you to reference a header/library/framework relative to somewhere in your search path. So in your case you would use `-F./Frameworks -framework MyFramework`. Make sure you’re adding this to both CFLAGS and LDFLAGS. LDFLAGS are only passed when linking, which would cause the first error you saw.

liuxuan30:
@kirb  thanks for this super fast reply!

I got another question, it seems myTweak_EXTRA_FRAMEWORKS feature are not working?

I put it under thoes/lib/ but when compiling, it either complains about header not found, or if I remove the code, just compile ,it says framework not found

liuxuan30:
OK I found the reason.

I'm putting it into theos/lib/, rather than theos/vendor/lib/, so it's missing...

@kirb I guess we need to update the wiki, kind of confusing it's using $THEOS/lib

PR #431, opened 2019-08-13
fix 3rd party framework path comment

liuxuan30:
Fix the comment, sorry but I cannot find the wiki file to update the path there

#430


liuxuan30:
I reviewed the makefile again, it seems the 'vendor' is not very clear to me

If I use XXX_EXTRA_FRAMEWORK flag, it seems checking theos/lib, while if use XXX_FRAMEWORK, it's just reading theos/vendor/lib, so @kirb looks like my PR is wrong?

can you identify XXX_EXTRA_FRAMEWORK and XXX_FRAMEWORK difference?

Is XXX_EXTRA_FRAMEWORK reads theos/lib, while XXX_FRAMEWORK is reading theos/vendor/lib?

kirb:
“Extra” frameworks are ones either bundled with Theos in `$THEOS/vendor/lib`, or installed by the user in `$THEOS/lib`. The comment saying `$THEOS/lib` probably predates us introducing the `vendor` directory.

I’ll consider this a wontfix.

Issue #400, opened 2019-04-07
Tweaks built for tvOS reference /Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate causing them not to load


Diatrus:
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

What are the steps to reproduce this issue?
-------------------------------------------
1. Build a tweak for tvOS and attempt to have it inject.

What happens?
-------------
Tweak will not inject and console will throw an error similar to
`Injection failed: 'dlopen(/usr/lib/TweakInject/WiivampTV.dylib, 9): Library not loaded: /Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate
  Referenced from: /usr/lib/TweakInject/WiivampTV.dylib
  Reason: image not found'`

What were you expecting to happen?
----------------------------------
Tweak should inject and not reference CydiaSubstrate, as nito himself has stated /Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate should not, and will not be symlinked with his jailbreak.

Any logs, error output, etc?
----------------------------
Maybe relevant: `ld: warning: invalid -install_name (/Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate) in dependent dylib (/Users/diatrus/theos/vendor/lib/CydiaSubstrate.framework/CydiaSubstrate.tbd). Dylibs/frameworks which might go in dyld shared cache cannot link with dylibs that won't be in the shared cache`

Any other comments?
-------------------
Nito himself has stated /Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate should not, and will not be symlinked to /usr/lib/libsubstrate.dylib with his jailbreak, and this is intended behavior.

What versions of software are you using?
----------------------------------------
**Operating System:** macOS 10.13.6

**SDK Version:** AppleTVOS11.3

uroboro:
Did you set the generator as `internal`?

Diatrus:
Same result with internal generator.

kirb:
I discussed this with @nitoTV, the framework symlink should be there but is missing by mistake. Till he fixes it, you can make it yourself: 

```shell
mkdir -p /Library/Frameworks/CydiaSubstrate.framework
ln -s /usr/lib/libsubstrate.dylib /Library/Frameworks/CydiaSubstrate.framework/CydiaSubstrate
```

Theos’s side is working as intended, so closing.

PR #436, opened 2019-08-24
Fix tweaks on AppleTV

Diatrus:
Modern AppleTV jailbreaks (>= AppleTV 4) by nitoTV haven't and never will (by his own admission) include CydiaSubstrate.framework. The default generator of MobileSubstrate links CydiaSubstrate.tbd to any tweak compiled for AppleTV, and this causes tweaks to not function, as they can't find the dynamically linked framework that isn't present on the device.

No, but would resolve a closed one. https://github.com/theos/theos/issues/400

N/A

Any other comments?
-------------------
While @kirb stated that he discussed it with nitotv and was told it was missing by mistake, multiple jailbreak updates have been released since and the framework is still not included. This, and him telling me himself it isn't meant to be there, leads me to believe it will never be fixed on the jailbreak's end. This PR fixes the issue and actually allows for tweaks built with upstream theos to work.

Where has this been tested?
---------------------------
**Operating System:** MacOS 10.13.6 & 10.15 beta 8

**Platform:** Darwin

**Target Platform:** tvOS

**Toolchain Version:** XCode 10.1, 11 beta 5

**SDK Version:** 10.1, 11.3, 12.4

### DIFF
--- a/makefiles/targets/Darwin/appletv.mk
+++ b/makefiles/targets/Darwin/appletv.mk
@@ -11,6 +11,7 @@ _THEOS_DARWIN_CAN_USE_MODULES := $(_THEOS_TRUE)
 NEUTRAL_ARCH := arm64
 
 _THEOS_TARGET_DEFAULT_OS_DEPLOYMENT_VERSION := 9.0
+_THEOS_TARGET_LOGOS_DEFAULT_GENERATOR := internal
 
 include $(THEOS_MAKE_PATH)/targets/_common/darwin_head.mk
 include $(THEOS_MAKE_PATH)/targets/_common/darwin_tail.mk


kirb:
I’d still really prefer against using the internal generator as much as possible. Does `libsubstrate.dylib` exist - just not `CydiaSubstrate.framework`? We can have it link the lib instead of the framework. However having one and not the other is still just annoying, especially if for instance I wanted to release a universal build of Cephei that runs on both iOS and tvOS via the same arm64 arch slice.

Diatrus:
@kirb ironically after sending this PR to nito, he told me he was just about to include that framework in the bootstrap after 5 months... No longer needed.

kirb:
Sounds good to me 😛

jamep09:
What good repos for nito tv

PR #462, opened 2019-11-19
Add xcodeproj support

kabiroberai:

This PR adds support for building Xcode projects while taking advantage of the rest of Theos' build system (eg. packaging, installation, subprojects). The primary use case is probably jailbroken apps. Sileo, Supercharge, and NewTerm all either use or have been tested with this branch.

Not that I know of

Nope

Any other comments?
-------------------
The one issue that I currently see is the fact that Xcode build failures don't propagate to the rest of Theos – i.e. subsequent tasks will continue without knowing that the build has failed. We should probably fix this before merging, which is why this is currently a draft PR.

Where has this been tested?
---------------------------
**Operating System:** macOS 10.14.6 (18G103)

**Platform:** macOS

**Target Platform:** iOS

**Toolchain Version:** Latest

**SDK Version:** Latest

### DIFF
--- a/makefiles/common.mk
+++ b/makefiles/common.mk
@@ -44,7 +44,7 @@ endif
 ifeq ($(_THEOS_INTERNAL_TRUE_PATH),)
 _THEOS_RELATIVE_MAKE_PATH := $(dir $(lastword $(MAKEFILE_LIST)))
 _THEOS_INTERNAL_TRUE_PATH := $(call __clean_pwd,$(_THEOS_RELATIVE_MAKE_PATH)/..)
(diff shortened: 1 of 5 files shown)


kirb:
Merged into [next](https://github.com/theos/theos/compare/master...next) branch. Needs testing. In particular I think we may need to implement bundle-level signing, which is supported by codesign and more recent versions of ldid. Seeing a few strange things with NewTerm (which is now being built with this) which I’m assuming must be codesigning-related.

kabiroberai:
@kirb we should close this PR and create an issue instead

kabiroberai:
Closing in favor of #489

Issue #482, opened 2020-02-12
arm64e compiliation on Linux


cbyrne:
Is there any updates for arm64e compiling on Linux?
The error: ```clang: error: invalid arch name '-arch arm64e'``` occurs when trying to compile, using an iOS 12.2 sdk. Guessing the toolchain needs to be updated, is this planned?

**Edit:**
After trying @sbingner's suggestion:
> You could try the toolchain I have at https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1

I now get the error: ``ld: unknown/unsupported architecture name for: -arch arm64e``. Then when using an iOS 12.4 sdk it can't build any modules i.e. ``CoreFoundation``, ``Foundation`` and ``Darwin``. Full log of this issue available here: https://psty.io/p?q=1ea6a

sbingner:
You could try the toolchain I have at https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1

ConorTheDev:
Will have a look at it

ConorTheDev:
Added ```~/ios-arm64e-clang-toolchain/bin/`` to my PATH variable. Getting a different error this time but no further compilation, does it need a newer SDK? Currently using 11.2, fyi.
![image](https://user-images.githubusercontent.com/31479541/74571897-33157180-4f76-11ea-9093-bd3724f4deff.png)

ConorTheDev:
Note: when using an iOS 12.4 patched SDK, I'm getting a massive console log of errors like this, theos on linux does need some work! :stuck_out_tongue: 
![image](https://user-images.githubusercontent.com/31479541/74572180-9ef7da00-4f76-11ea-90af-e69f3cbc19dd.png)

sbingner:
you probably need to use wrapper scripts to execute clang with -target etc properly

uroboro:
Fixed with 8771215f765e774bfefdacc6ddf0e151c2973d49

LaxusiC:
@sbingner I'm new to making tweak, how do I install the toolchain correctly? I extract and put in /usr/ but still can't compile for arm64e. Am I doing wrong?

PR #499, opened 2020-04-17
Remove $(SDKTARGET)- prefix name requirement for linux toolchain binaries

DGh0st:
Removes the need to have prefixes, such as `armv7-apple-darwin-` and `arm64-apple-darwin-` for the linux toolchains. With this change it should make sbingner's arm64e linux toolchain ([https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1](https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1)) "plug and play".

#482 

Any other comments?
-------------------
This is going to break backwards compatibility with old and current Kabir's linux toolchain which has those prefixes.

Where has this been tested?
---------------------------
**Operating System:**
- Windows 10 (Debian WSL)
- Debian

**Platform:**
- Linux

**Target Platform:**
- iPhone

**Toolchain Version:**
- [https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1](https://github.com/sbingner/llvm-project/releases/tag/v10.0.0-1)

**SDK Version:**
- 13.0 (Debian WSL)
- 11.2 (Debian)

### DIFF
--- a/makefiles/targets/Linux/iphone.mk
+++ b/makefiles/targets/Linux/iphone.mk
@@ -11,7 +11,7 @@ SWIFTBINPATH ?= $(THEOS)/toolchain/swift/bin
 SDKBINPATH ?= $(THEOS)/toolchain/$(THEOS_PLATFORM_NAME)/$(THEOS_TARGET_NAME)/bin
 
 # Determine toolchain to use based on file existence.
-ifeq ($(SDKTARGET),)
+ifeq ($(SDKTARGETPREFIX),)
 ifeq ($(call __exists, $(SDKBINPATH)/armv7-apple-darwin11-ld),$(_THEOS_TRUE))
 SDKTARGETPREFIX ?= armv7-apple-darwin11-
 else ifeq ($(call __exists, $(SDKBINPATH)/arm64-apple-darwin14-ld),$(_THEOS_TRUE))
--- a/makefiles/targets/Linux/iphone.mk
+++ b/makefiles/targets/Linux/iphone.mk
@@ -17,7 +17,7 @@ SDKTARGETPREFIX ?= armv7-apple-darwin11-
 else ifeq ($(call __exists, $(SDKBINPATH)/arm64-apple-darwin14-ld),$(_THEOS_TRUE))
 SDKTARGETPREFIX ?= arm64-apple-darwin14-
 else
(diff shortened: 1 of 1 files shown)


makefiles/targets/Linux/iphone.mk
```diff
@@ -7,20 +7,28 @@ _THEOS_TARGET_PLATFORM_SDK_NAME := iPhoneOS
...
-ifeq ($(wildcard $(THEOS)/toolchain/$(THEOS_PLATFORM_NAME)/$(THEOS_TARGET_NAME)/bin/arm64-apple-darwin14-ld),)
-SDKTARGET ?= armv7-apple-darwin11
+ifeq ($(SDKTARGETPREFIX),)
+ifeq ($(call __exists, $(SDKBINPATH)/armv7-apple-darwin11-ld),$(_THEOS_TRUE))
+SDKTARGETPREFIX ?= armv7-apple-darwin11-
+else ifeq ($(call __exists, $(SDKBINPATH)/arm64-apple-darwin14-ld),$(_THEOS_TRUE))
+SDKTARGETPREFIX ?= arm64-apple-darwin14-
 else
-SDKTARGET ?= arm64-apple-darwin14
+SDKTARGETPREFIX ?=
```
DGh0st, line 20: I just realized this assignment is pretty pointless and the check on L14 is going to always be true when it doesn't have a prefix so its not adding much.

DGh0st: Removed this pointless assignment.

I left the L14 check in to keep same behavior for current toolchains. Not sure how that performance gain can be added for the toolchains with no prefix.

DGh0st:
> Additionally, is there a reason that L31-L33 are inserted _after_ the includes?

`darwin_tail.mk` is where those 3 variables are initially setup (if not already set) and then eventually used in `rules.mk`. So setting it before the include would lose all the flags that would have been setup in `darwin_tail.mk`.

> Most of Theos uses `_THEOS_<component>_` prefixes for variables. In this case, I'd suggest at least changing `TARGET_OPTIONS` to `_THEOS_TARGET_TRIPLE_FLAG`.

Yeah I can rename that and `SDKTARGETPREFIX` to match that style.

Issue #515, opened 2020-05-27
`INSTALL_TARGET_PROCESSES` doesn't unload processes on macOS


jacobcxdev:
`INSTALL_TARGET_PROCESSES` doesn't unload processes when installing SIMBL tweaks with `make do`.

What are the steps to reproduce this issue?
-------------------------------------------
1. Create SIMBL tweak
2. make do

What happens?
-------------
The processes aren't unloaded.

What were you expecting to happen?
----------------------------------
The processes to be unloaded.

Any logs, error output, etc?
----------------------------
Nope.

Any other comments?
-------------------
Nope.

What versions of software are you using?
----------------------------------------
**Operating System:** 10.15.4

**Toolchain Version:** …

**SDK Version:** 10.15.4

uroboro:
Does this happen with any process?

jacobcxdev:
> Does this happen with any process?


@uroboro Yep. I’m pretty sure it was happening because of one of the errors I was getting, because it was magically fixed in #516.

Issue #514, opened 2020-05-27
`su: illegal option -- c` when installing SIMBL tweak


jacobcxdev:
Theos produces the error `su: illegal option -- c` when installing a SIMBL tweak. I'm assuming it's because of `_THEOS_SUDO_COMMAND` being incorrect, but don't take my word for it.

What are the steps to reproduce this issue?
-------------------------------------------
1. Create a SIMBL tweak
2. `make do`

What happens?
-------------
```
==> Installing…
su: illegal option -- c
usage: su [-] [-flm] [login [args]]
make: *** [/Users/jacob/.theos/makefiles/install/pkg_local.mk:2: internal-install] Error 1
```

What were you expecting to happen?
----------------------------------
Successful installation.

Any logs, error output, etc?
----------------------------
```
==> Installing…
su: illegal option -- c
usage: su [-] [-flm] [login [args]]
make: *** [/Users/jacob/.theos/makefiles/install/pkg_local.mk:2: internal-install] Error 1
```

Any other comments?
-------------------
Nope.

What versions of software are you using?
----------------------------------------
**Operating System:** macOS 10.15.4

**Toolchain Version:** …

**SDK Version:** 10.15.4

uroboro:
A quick inspection of our sudo usage, we use `sudo`, or not at all if the user is already root. Can you check what your `sudo` command is? It could be a shell script doing `su -c $@` and that might cause this issue.

uroboro:
I caught up to #516 which clearly shows that `install.exec` is what does what I suggested

PR #516, opened 2020-05-27
Fixed installing + uninstalling SIMBL tweaks via `make do`, `make install` and `make uninstall`.

jacobcxdev:
Fixed installing + uninstalling SIMBL tweaks via `make do`, `make install` and `make uninstall`.

Fixed installing + uninstalling SIMBL tweaks via `make do`, `make install` and `make uninstall`.
- Updated default install path to MacForge's plugin directory
- Fixed up uninstall command to not break when encountering spaces in file paths (and silenced the `rm -d` output)
- Fixed the local install command in `bin/install.exec` (thanks @kirb).

- #514 
- #515 

Nope.

Any other comments?
-------------------
mySIMBL is pretty old now, and MacForge has become the standard injection program for macOS. IMO it's time to retire mySIMBL support in favour of MacForge support, which is exactly what this PR does.

Where has this been tested?
---------------------------
**Operating System:** macOS 10.15.4

**Platform:** macOS

**Target Platform:** macOS

**Toolchain Version:** …

**SDK Version:** 10.15.4

### DIFF
--- a/bin/install.exec
+++ b/bin/install.exec
@@ -15,5 +15,5 @@ if [[ $TARGET_INSTALL_REMOTE == 1 ]]; then
 
 	exec ${args[@]}
 else
-	exec su -c "$@"
+	exec bash -c "$@"
 fi
--- a/makefiles/install/pkg_local.mk
+++ b/makefiles/install/pkg_local.mk
@@ -2,4 +2,4 @@ internal-install:: internal-install-check
 	$(ECHO_INSTALLING)install.exec "$(_THEOS_SUDO_COMMAND) installer -pkg \"$(_THEOS_PACKAGE_LAST_FILENAME)\" -target /"$(ECHO_END)
 
 internal-uninstall::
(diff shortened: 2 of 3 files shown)


kirb:
This looks good to merge aside from the install.exec change - I’ll need to have a think of how this can work in all cases. I think iOS running as mobile is the only case where we do want to use su. (Would rather this be passed through from the Theos platform makefile than just checking `[[ $USER == mobile ]]`)

jacobcxdev:
> This looks good to merge aside from the install.exec change - I’ll need to have a think of how this can work in all cases. I think iOS running as mobile is the only case where we do want to use su. (Would rather this be passed through from the Theos platform makefile than just checking `[[ $USER == mobile ]]`)

That would indeed be preferable. I considered checking whether `[[ $(uname -p) == *"arm"* ]]` (I think that works? It’s slightly more solid than checking $USER), but in the end I decided to leave it for you to decide how to handle!

Diatrus:
Might I also comment that iOS jailbreaks utilizing BSD su do not include the `-c` option either, instead allowing it’s use as a shell argument. E.g. `su root -c whoami` would output “root”. A possible solution, in this case, would be checking if the sudo command exists and using it in lieu of su.

kirb:
Ah damn, this got closed because I merged next. Can you change the destination branch to master and reopen?

PR #573, opened 2021-02-13
Speed improvements

sbingner:

This increases the speed of loading makefiles.  I got it down to about 550ms from 1050ms on my system.

I expect this will take some discussion :)


makefiles/common.mk
```diff
@@ -34,6 +36,96 @@ __executable = $(if $(shell PATH="$(THEOS_BIN_PATH):$$PATH" type "$(1)" > /dev/n
 __simplify = $(2)$(eval $(1):=$(2))
 ###
 
+empty :=
+space := $(empty) $(empty)
+
+define __cmp
```
kirb, line 42: Haha, wow, that’s definitely one hell of a solution! Well done.

My comments would be:

1. Let’s move this to another file. As it’s very complex, it interrupts the flow of common.mk otherwise.
2. If it isn’t too crazy to, there should be comments explaining what each line is doing - inline with the logic if Make is cool with it, but I bet it won’t be, so just above would be fine.
3. Add another underscore prefix to the “internal” helper functions of the “public” __vercmp function.

kirb (changes requested):

makefiles/common.mk
sbingner: OK, I think I handled those issues... and removed the erroneously included submodule-commit changes :|

kirb (approved):

kirb:
Squashed the vercmp commits and merged in 6ddf05e and a9760e6.

PR #595, opened 2021-05-27
[Linux/iphone] Fix -target flag being overwritten

sbingner:

theos currently ignores `target=` commands for linux because it sets `-target arm64-apple-darwin` unconditionally after whatever the correct `-target` command is and no longer passes legacy flags

-----------------------------------------
Not that I know of


kirb (approved):

UsrLightmann:
Can confirm that w/ these changes,  `-target` commands now work as expected.

PR #596, opened 2021-05-29
Fix TARGET_OS_IPHONE not being defined in Prefix.pch (Resolves #552)

### LINKED ISSUE #552: The latest Prefix.pch breaks projects
What are the steps to reproduce this issue?
-------------------------------------------
1. Build a tweak project using latest Theos

What happens?
-------------
Report some errors about missing headers

What were you expecting to happen?
----------------------------------
Successfully build the project

Any logs, error output, etc?
----------------------------
None

Any other comments?
-------------------
After digging in for some while, I found that the new Prefix.pch uses the macro TARGET_OS_IPHONE instead of TARGET_IPHONE and on my machine it is not defined. I also found that the common.mk and iphone.mk defines -DTARGET_IPHONE=1, so I added -DTARGET_OS_IPHONE=1 in cflags and it works. Is it a problem only for my environment or it is a common problem?

What versions of software are you using?
----------------------------------------
**Operating System:** MacOS 10.15

**Toolchain Version:** From Xcode 11.7

**SDK Version:** iPhone 11.2

opa334:
TargetConditionals.h needs to be included in Prefix.pch, otherwise TARGET_OS_IPHONE might not be defined which causes the later checks to fail.

Yes, #552 

See the issue for more information

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** macOS 11.4

**Platform:** macOS

**Target Platform:** iOS

**Toolchain Version:** Xcode 12.5

**SDK Version:** iPhoneOS13.0

### DIFF
--- a/Prefix.pch
+++ b/Prefix.pch
@@ -10,6 +10,10 @@
 // has already been removed for other platforms, such as macOS and tvOS, and is not used
 // at all for Swift files.
 
+#if __has_include(<TargetConditionals.h>)
+#include <TargetConditionals.h>
+#endif
+
 #ifdef DEBUG
 	#define __DEBUG__
 #endif


PoomSmart (approved):

kabiroberai:
Sorry for getting to this a bit late but I believe this PR has some semantics-breaking ramifications; the `!defined(__IPHONE_14_0) && !defined(THEOS_LEAN_AND_MEAN)` check should probably be moved higher up in order to avoid importing TargetConditionals.h unless those conditions are satisfied. Otherwise, we're breaking the guarantee of the header being a no-op in those two conditions.

PoomSmart:
cc @opa334

opa334:
> Sorry for getting to this a bit late but I believe this PR has some semantics-breaking ramifications; the `!defined(__IPHONE_14_0) && !defined(THEOS_LEAN_AND_MEAN)` check should probably be moved higher up in order to avoid importing TargetConditionals.h unless those conditions are satisfied. Otherwise, we're breaking the guarantee of the header being a no-op in those two conditions.

I agree, however it seems like _THEOS_IS_SIMULATOR depends on TARGET_OS_IPHONE and has to be defined even if THEOS_LEAN_AND_MEAN is set, which kinda creates a chicken egg problem as far as I can see.

kabiroberai:
From the looks of it, `_THEOS_IS_SIMULATOR` is a private define that's only really intended for use inside the prefix header itself. I think it's okay if it's not present under THEOS_LEAN_AND_MEAN (in fact it seems like it was supposed to be `#undef`'d after being used, but along the way some variable names were updated and that stopped being the case).

PR #610, opened 2021-07-04
Use git URLs instead of HTTPS URLs to work around unsupported TLS version on iOS 5.1.1

daviewales:

Use git URLs instead of HTTPS URLs to work around unsupported TLS version on iOS 5.1.1

GitHub won't accept requests if the TLS version is too low.
If you try to clone theos using https on an old iOS version such as iOS 5.1.1 which doesn't support modern TLS, it gets stuck when it tries to clone the submodules:

e.g. An example of the error when you try to clone using HTTPS on iOS 5.1.1
```
fatal: unable to access 'https://github.com/theos/theos.git/': error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

This change updates the submodules to use `git://` URLs rather than `https://` URLs, which allows the following to just work on iOS 5.1.1:

```
 git clone --recursive git://github.com/theos/theos.git $THEOS
```

The wiki will also need to be updated to change git clone command to the above if this PR is accepted.

Where has this been tested?
---------------------------
**Operating System:** iOS 5.1.1

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,18 +1,18 @@
 [submodule "vendor/dm.pl"]
 	path = vendor/dm.pl
-	url = https://github.com/theos/dm.pl.git
+	url = git://github.com/theos/dm.pl.git
 [submodule "vendor/include"]
 	path = vendor/include
-	url = https://github.com/theos/headers.git
+	url = git://github.com/theos/headers.git
 [submodule "vendor/lib"]
 	path = vendor/lib
-	url = https://github.com/theos/lib.git
+	url = git://github.com/theos/lib.git
 [submodule "vendor/logos"]
 	path = vendor/logos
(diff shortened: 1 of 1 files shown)


kirb:
Unfortunately this isn’t something we can merge, because git:// is unencrypted, and it’s not worth the compromise to change this for an old platform. I believe you can install the CA Certs (org.thebigboss.cacerts) package to fix this, but in general using old iOS as a dev environment just doesn’t seem like a good idea as the latest toolchain that works on Telesphoreo bootstrap (i.e. iOS 2 - 11) is pretty old.

daviewales:
Thanks, I thought that might be the case. I've got CA Certs, but I think the issue is due to a low TLS version, which I haven't found a fix for anywhere.

Issue #566, opened 2021-01-09
Update Wiki to Use New toolchain


CRKatri:
[I have recently released a new arm64e toolchain for use with theos](https://github.com/CRKatri/llvm-project/releases/tag/swift-5.3.2-RELEASE), it is available for Debian Buster, Ubuntu 18.04, and Ubuntu 20.04, although the ubuntu toolchains should work on a wide variety of distros. They contain updated cctools and libtapi. It is compatible for both objc and swift development. This will allow swift development to be done from Linux.

It's tools are updated from @sbingner's arm64e llvm/clang toolchain and @kabiroberai's swift 4 toolchain.

kabiroberai:
There's two approaches to be considered here:

1. Continue keeping the Clang and Swift toolchains separate, since that way users can save space if they don't plan on using Swift. At the moment, `clang` is still being duplicated because swiftc comes with its own copy, but this could be remedied by specifying `-tools-directory` to swiftc and pointing it at the other toolchain.
2. Use a single toolchain with everything in it. This has the benefit of not needing to worry about Swift possibly using a different Clang, and can also make sure that both Clang and Swift are compiled with the same llvm version etc so that there isn't any discrepancy in, say, the supported tapi version.

I'm not sure which is better tbh, but I'm interested in hearing what others have to say about this.

PR #621, opened 2021-09-03
Swift tweak development with Orion + significant Swift improvements

### LINKED ISSUE #456: THEOS ld: warning
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

What are the steps to reproduce this issue?
-------------------------------------------
1. … Install Theos on Mac OS 10.15 Catalina
2. … Install latest command line tools (11.0.0.0.1.1567737322)
3. … Run Make on any template

What happens?
-------------
…
make successful with errors

What were you expecting to happen?
----------------------------------
…
no errors

Any logs, error output, etc?
----------------------------
ld: warning: building for iOS, but linking in .tbd file (/Users/USER/theos/vendor/lib/CydiaSubstrate.framework/CydiaSubstrate.tbd) built for iOS Simulator

Any other comments?
-------------------
…this happened after updating from Mojave > Catalina

What versions of software are you using?
----------------------------------------
**Operating System:** … Mac OS 10.15 Catalina

**Toolchain Version:** … Xcode 11 (CLTools 11.0.0.0.1.1567737322)

**SDK Version:** … 11.2

### LINKED ISSUE #500: xcodeproj build output not streamed during parallel builds
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

What are the steps to reproduce this issue?
-------------------------------------------
1. Create an xcodeproj-based Theos project
2. Run `make` using GNU Make 4

What happens?
-------------
There is no output until the very end, at which point all of it appears at once. Additionally, there's a warning that says `-j8 forced in makefile: resetting jobserver mode.`

What were you expecting to happen?
----------------------------------
The output should be streamed.

Any logs, error output, etc?
----------------------------
Example using `make clean`:
```shell
Kabirs-MacBook-Pro:Supercharge kabiroberai$ make clean
# [note: there's a long wait here]
==> Cleaning…
> Making clean in superchargesb…
make[1]: warning: -j8 forced in makefile: resetting jobserver mode.
==> Cleaning…
> Making clean in scfileops…
make[1]: warning: -j8 forced in makefile: resetting jobserver mode.
==> Cleaning…
```

Any other comments?
-------------------
It appears that this can be resolved by forcing a non-parallel build `-j1` (eg `make -j1`). I did try `--output-sync=none` with a parallel build, however that didn't fix the issue. Forcing non-parallel xcodeproj builds might actually be a viable solution, since afaik Theos doesn't parallelize `xcodebuild` invocations anyway.

What versions of software are you using?
----------------------------------------
**Operating System:** macOS 10.14.6 (18G103)

**Toolchain Version:** Xcode 11.2.1 (11B53)

**SDK Version:** iOS 13.2.2

### LINKED ISSUE #504: Hooks in Swift
Hi, is there any way to hook stuff, and make a tweak entirely in swift with no objective C. 

I've been [playing around with AutoHook](https://github.com/level3tjg/AutoHook-Swift) but it's super slow and causes performance issues even when you only have one or two hooks.

### LINKED ISSUE #524: Theos trying to link phantom object files
What are the steps to reproduce this issue?
-------------------------------------------
1. [Download this sample project](https://github.com/theos/theos/files/4805235/Test.zip)
2. Run `make` in the root project dir

What happens?
-------------
Theos tries to link a phantom object file.

What were you expecting to happen?
----------------------------------
Either Theos to link an existing object file, or to generate the phantom object file.

Any logs, error output, etc?
----------------------------
https://paste.mod.gg/ibukeyoyaf.coffeescript

Any other comments?
-------------------
This is a really hacky 'solution' to trying to share a Swift class between a project and subprojects.

What versions of software are you using?
----------------------------------------
**Operating System:** macOS 10.15.5

**Toolchain Version:** …

**SDK Version:** iOS 13.5

kabiroberai:
(Reopened based on the renamed `orion` branch instead of `kabiroberai/orion`)

**Note**: ~~As of the creation of this PR, the [theos/orion](https://github.com/theos/orion) repository at the core of this PR is currently private. It will be made public once Orion is deemed ready for public beta testing. While you probably can't use Orion yet,~~ The Orion repo is now public; feel free to check out the preliminary documentation at https://orion.theos.dev to acquaint yourself with the project.

This pull request:
- Adds support for tweak development in Swift using [Orion](https://orion.theos.dev)
- Adds a novel way to work on Theos projects in any LSP-compatible editor (even on Windows and Linux!) with full autocomplete support by using a Swift Package to mirror the project's source
- Improves Clang module support: you can now import anything from [vendor/]lib or [vendor/]include (in both ObjC as well as Swift) without needing to use the `instance_FRAMEWORKS` variable as long as modules are enabled. In addition, some vendored libraries have been adorned with module support which makes them possible to use in Swift and easier to use in Objective-C.
- Adds tweak_swift and preference_bundle_swift templates.
- Changes our approach with Swift so that projects are built the "correct" way – by invoking the Swift driver instead of directly calling the frontend, the latter of which has an unstable API. This is achieved by using a few support tools written in Swift, bootstrapped with SPM. This will also enable us to write additional tools using Swift in the future, if Make/Perl aren't sufficient. For instance, this will make it easier to eventually add SPM dependency support to Theos projects as we will be able to rely on libSwiftPM.

To my knowledge:
Closes #456
Closes #500
Closes #504 
Closes #524

Nope

Any other comments?
-------------------
This shouldn't be merged until Orion is beta-tested to ensure that it works, because once Orion lands in master the ABI should effectively be considered frozen and any further changes can only be additive.

Where has this been tested?
---------------------------
**Operating System:** macOS 11.1 (20C69), Ubuntu 18.04

**Platform:** macOS, Linux

**Target Platform:** macOS (needs more testing), iOS

**Toolchain Version:** Xcode 11.7, Xcode 12.3, my Swift 5.3 Linux [toolchain](https://github.com/kabiroberai/swift-toolchain-linux/tree/swift-5-support) + sbingner's toolchain, CRKatri's Swift 5.3 Linux toolchain (#566)

**SDK Version:** iOS 13.7, iOS 14.3, macOS 11.1


makefiles/common.mk
```diff
@@ -192,16 +202,17 @@ SHOULD_STRIP := $(call __theos_bool,$(or $(strip),$(STRIP),$(_THEOS_SHOULD_STRIP
 
 ifeq ($(SHOULD_STRIP),$(_THEOS_TRUE))
 OPTFLAG ?= -Os
-SWIFT_OPTFLAG ?= -O -whole-module-optimization
+SWIFT_OPTFLAG ?= -O -whole-module-optimization -num-threads 1
```
L1ghtmann, line 205: Is there a reason we want to run single threaded here?
```suggestion
SWIFT_OPTFLAG ?= -O -whole-module-optimization -num-threads $(shell $(_THEOS_PLATFORM_GET_LOGICAL_CORES))
```

makefiles/instance/rules.mk
```diff
@@ -124,6 +134,24 @@ else
...
 
+_TARGET_ARCHS_COUNT = $(words $(TARGET_ARCHS))
+
+# The magic behind `make commands`: proxy the compile commands for one arch
+# through gen-commands.pl so that we can save them to the commands db
+export THEOS_GEN_COMPILE_COMMANDS
+ifeq ($(THEOS_GEN_COMPILE_COMMANDS) $(THEOS_CURRENT_ARCH),$(_THEOS_TRUE) $(firstword $(TARGET_ARCHS)))
+_TARGET_SWIFTC := $(TARGET_SWIFTC)
+# we use PWD instead of THEOS_PROJECT_DIR because the latter always refers to the root level project
+# so it isn't correct for subprojects.
```
L1ghtmann, line 145: Do we need the `TARGET_SWIFTC` and `TARGET_CXX` vars to be re-evaluated each call? If not, we might want to consider changing them to simple assigns (`:=`) so that we can do `TARGET_CXX := <more stuff here> $(TARGET_CXX)`  or `TARGET_CXX := $(addprefix <some text>, $(TARGET_CXX))`

makefiles/master/rules.mk
```diff
@@ -184,6 +217,25 @@ else
 	$(ERROR_BEGIN) "You don't have the GitHub CLI installed. For more information, refer to: https://cli.github.com/" $(ERROR_END)
 endif
 
+# The SPM config is a simple key-value file used to pass build settings to Package.swift.
+# Each line is either empty or starts with a (unique) key, followed by an equals sign, 
+# followed by the key's value.
+spm::
+	@$(PRINT_FORMAT_MAKING) "Creating SPM config (beta)"
```
L1ghtmann, line 224: Are we still considering this feature a "beta?"

makefiles/platform/Darwin.mk
```diff
@@ -41,4 +41,5 @@ endif
 
 THEOS_SUDO_COMMAND ?= sudo
 _THEOS_PLATFORM_GET_LOGICAL_CORES := sysctl -n hw.ncpu
+_THEOS_PLATFORM_SWIFT ?= swift
```
L1ghtmann, line 44: Is there a reason we're allowing reassignment here but not on a Linux host?

makefiles/targets/_common/darwin_head.mk
```diff
@@ -60,25 +55,46 @@ else
...
 
 TARGET_STRIP_FLAGS ?= -x
 
 ifeq ($(TARGET_DSYMUTIL),)
-ifeq ($(call __executable,$(call __invocation,dsymutil)),$(_THEOS_TRUE))
-	TARGET_DSYMUTIL = $(call __invocation,dsymutil)
-else ifeq ($(call __executable,$(call __invocation,llvm-dsymutil)),$(_THEOS_TRUE))
-	TARGET_DSYMUTIL = $(call __invocation,llvm-dsymutil)
-endif
+	TARGET_DSYMUTIL := $(call __invocation,dsymutil)
```
L1ghtmann, line 91: Since `dysmutil` is vended by both Xcode and the Linux/iOS toolchains, would it be worth assigning `TARGET_DSYMUTIL` along with the other TARGETs above and reassigning only if it's blank (for whatever reason)?

L1ghtmann:
Have a handful of questions regarding minor things but otherwise this looks great! Can't personally speak to the provided functionality, as I don't know or write Swift, but I can't recall hearing any pressing issues following the many months of open testing.

makefiles/messages.mk
```diff
@@ -56,17 +56,16 @@ else
 endif
 
 	ECHO_PREPROCESSING = $(ECHO_BEGIN)$(PRINT_FORMAT_RED) "Preprocessing $<"$(ECHO_END); $(ECHO_PIPEFAIL) (
+	ECHO_PREPROCESSING_XSWIFT = $(ECHO_BEGIN)$(PRINT_FORMAT_RED) "Preprocessing $(subst $(_SPACE),$(_COMMA) ,$^)"$(ECHO_END); $(ECHO_PIPEFAIL) (
```
L1ghtmann, line 59: We may be able to make use of `$(foreach` (assuming `$^` is a space-separated list) to avoid having to define `_SPACE` and `_COMMA`.
```suggestion
	ECHO_PREPROCESSING_XSWIFT = $(ECHO_BEGIN)$(PRINT_FORMAT_RED) "Preprocessing $(foreach x,$^,$(x),)"$(ECHO_END); $(ECHO_PIPEFAIL) (
```

makefiles/common.mk
kabiroberai: iirc Xcode uses one thread for release builds — I'm not sure what the rationale is but I assume there's a build time vs runtime tradeoff here.

makefiles/messages.mk
kabiroberai: This would add a trailing comma after the last filename, wouldn't it?

makefiles/platform/Darwin.mk
kabiroberai: Not particularly, though this is underscored so clients aren't expected to override it in the first place. I do see the value in reassigning it though, so I'd be happy to add a configurable non-underscored variant — but I might save that for a follow-up PR.

makefiles/targets/_common/darwin_head.mk
kabiroberai: This is one of those cases where the answer is "that sounds reasonable" but due to the non-standard `llvm-dsymutil` checking going on here I suspect that there's a reason for structuring the code this way; and imo when it comes to GNU Make if it ain't broke don't fix it

makefiles/master/rules.mk
kabiroberai: Good point, might be time for this to graduate :)

makefiles/common.mk
absidue: It's because of optimisations. If you compile single threaded, optimisations can be done on everything. If you compile multi-threaded, optimisations are done in each thread/compilation unit separately, so you miss out on opportunities to inline functions and eliminate dead code, because the threads don't know what the other threads are doing.

makefiles/master/rules.mk
```diff
@@ -18,18 +21,43 @@ THEOS_USE_PARALLEL_BUILDING := $(_THEOS_IS_MAKE_GT_4_0)
...
+endif
+export _THEOS_INTERNAL_USE_PARALLEL_BUILDING
+
+# certain conditions need to execute, semantically, when we're
+# running the first `all`. This is usually when MAKELEVEL == 0
+# but is in fact MAKELEVEL == 1 if we're running `troubleshoot`
+_THEOS_TOP_ALL_MAKELEVEL := $(if $(THEOS_IS_TROUBLESHOOTING),1,0)
+ifeq ($(MAKELEVEL),$(_THEOS_TOP_ALL_MAKELEVEL))
+_THEOS_IS_TOP_ALL := $(_THEOS_TRUE)
+endif
```
kirb, line 36: Not sure if we should make this assumption in this way, since we can be called by another non-Theos makefile too. Maybe we could look for something we export later on?

kabiroberai: The true semantic intent here is to recreate the auxiliary/markers directories once at the start of a build — I wonder if there's a more Make-esque way to do this using some sort of rule/intermediate. IMO this is good enough for now but I can follow-up with a better approach when I have the chance.

PR #626, opened 2021-11-07
Disable Bitcode when using xcodeproj.mk

elihwyma:
When distributing apps anywhere other than the App Store the bitcode section of an app goes unused. This means it won't be stripped during processing on Apple's Servers. Disabling bitcode can make apps significantly smaller. Here is Zebra with Bitcode disabled (left) and enabled (right).
image

For further reference on Bitcode you can read [this](https://help.apple.com/xcode/mac/11.0/index.html?localePath=en.lproj#/devde46df08a) help article from Apple.


kabiroberai:
Good suggestion; can we make this configurable? The default being `NO` makes sense, but I can think of use cases where the user might want to override it, for example if you're using xcodeproj.mk to build an xcframework for distribution.

elihwyma:
That makes sense. On that same note it may also be worth making ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES a configurable option

elihwyma:
Finally got round to updating this, now by default Bitcode will be disabled however it can be enabled by doing 
`ExampleApp_ENABLE_BITCODE = YES`

PoomSmart (approved):

PR #634, opened 2022-01-16
Fix orion git submodule url

SwiftWinds:
Without this, `git submodule update --init --recursive` fails with `fatal: not a git repository: ../../.git/modules/vendor/orion`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,4 +21,4 @@
 	url = https://github.com/theos/swift-support.git
 [submodule "vendor/orion"]
 	path = vendor/orion
-	url = https://github.com/theos/orion
+	url = https://github.com/theos/orion.git


kirb:
Interesting, I think you’ve hit a Git bug? Can’t tell why .git/modules/vendor/orion would be broken otherwise. It might be that just changing the url made Git reset some internal state and bypass the bug, as in, adding .git wasn’t necessarily what fixed it. GitHub supports urls both with and without .git on the end, though it would still make sense to keep consistent and have all of our submodules use .git.

SwiftWinds:
Hmm yeah it's weird. Seems to work once I add the `.git` though ¯\_(ツ)_/¯

kabiroberai (approved):
I agree that using .git is good for consistency's sake. As for the root cause, I'm not too sure either -- @kirb's guess might be right.

PR #641, opened 2022-03-27
Change .s files to be treated as assembler-with-cpp

captinc:
This changes .s files to be treated as `assembler-with-cpp` instead of `assembler` ([see clang's -x option](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-x-language))

Note that the above 2 definitions are slightly misleading: 
* `assembler` actually means "treat as assembly source code _without_ using the C preprocessor"
* `assembler-with-cpp` actually means "treat as assembly source code _with_ using the C preprocessor"

Although this change is contrary to clang's defaults, Xcode does indeed treat .s as `assembler-with-cpp`, so I feel that Theos should match that

Not that I know of

N/A

Any other comments?
-------------------
None

Where has this been tested?
---------------------------
**Operating System:**
macOS Monterey 12.3 on M1

**Platform:**
macOS

**Target Platform:**
iOS

**Toolchain Version:**
Xcode Command Line Tools 2395

**SDK Version:**
iOS 14.5


kirb (approved):

kirb:
Merged, thanks!

PR #656, opened 2022-07-03
Handle attempting to update theos as root

### LINKED ISSUE #406: update-theos: Handle cases where $THEOS isn't writable
Background:
On macOS theos is usually installed to ``/opt/theos``, which is fine as long as you have sudo.

However, most naive users are calling tools from the non-sudoers user, this is entirely fine.

UNTIL:
The user decides to call ``sudo $THEOS/bin/update-theos``, since sudo doesn't pass environment variables by default as least on BSD/macOS, the user would be really lost at the cryptic message ``$THEOS must be set to the location of Theos to use ``. I suggest warning the user with a message like "If you are calling with sudo, remember to pass -E" to that error message.

I'm sure this has been asked a million times, this is just my 5 cents.

Zhang

L1ghtmann:
- Checks for root in `update-theos` and prompts the user to 1) not use root and 2) adjust their install location if that's the driving factor

It should resolve #406 

```
Theos should NOT be installed with or run as root (su/sudo)!
  - If you've installed Theos as root, please remove the install and reinstall as a non-root user.
  - If you're using root because you don't have write permission for Theos' parent directory, please either install Theos to a different directory where you do have write permission or, if you're trying to install to /opt, change the directory's ownership before attempting to install Theos as a non-root user.
```

Any other comments?
-------------------
I tried to make the error as helpful as possible, but if you'd like me to slim it down just say the word. 

Where has this been tested?
---------------------------
**Operating System:** …

iOS - 14.3 - Taurine  

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/update-theos
+++ b/bin/update-theos
@@ -5,22 +5,30 @@ has() {
 	type "$1" >/dev/null 2>&1
 }
 
+# Root is no bueno
+if [[ $EUID -eq 0 ]]; then
+	echo "Theos should NOT be installed with or run as root (su/sudo)!" >&2
+	echo "  - If you've installed Theos as root, please remove the install and reinstall as a non-root user." >&2
+	echo "  - If you're using root because you don't have write permission for Theos' parent directory, please either install Theos to a different directory where you do have write permission or, if you're trying to install to /opt, change the directory's ownership before attempting to install Theos as a non-root user." >&2
+	exit 1
+fi
+
 # Ensure $THEOS is set and is a directory.
 if [[ -z "$THEOS" || ! -d "$THEOS" ]]; then
 	echo "\$THEOS must be set to the location of Theos to use $(basename $0)." >&2
-	exit 1
+	exit 2
 fi
 
 # Ensure $THEOS is a Git repo.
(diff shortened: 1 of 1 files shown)


bin/update-theos
```diff
@@ -5,22 +5,30 @@ has() {
 	type "$1" >/dev/null 2>&1
 }
 
+# Root is no bueno
+if [[ $EUID -eq 0 ]]; then
+	echo "Theos should NOT be installed with or run as root (su/sudo)!" >&2
+	echo "  - If you've installed Theos as root, please remove the install and reinstall as a non-root user." >&2
+	echo "  - If you're using root because you don't have write permission for Theos' parent directory, please either install Theos to a different directory where you do have write permission or, if you're trying to install to /opt, change the directory's ownership before attempting to install Theos as a non-root user." >&2
```
kirb, line 12: This text looks good. Could you wrap it at 80 columns? Good to merge once that's done.

kirb (changes requested):

kirb:
Merged with the text wrapping sorted out, thanks!

PR #635, opened 2022-02-06
Add FreeBSD host platform and use /usr/bin/env in shebangs

CRKatri:
- Remove bashisms
- Add FreeBSD platform

What does this implement/fix? Explain your changes.
---------------------------------------------------
…

Does this close any currently open issues?
------------------------------------------
…

Any relevant logs, error output, etc?
-------------------------------------

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

Issue #616, opened 2021-07-17
[iOS] Theos Dependencies package and iOS Toolchain are outdated


UsrLightmann:
(Thanks for reporting an issue! Please make sure you click the link above to view the issue guidelines, then fill out the blanks below.)

What are the steps to reproduce this issue?
-------------------------------------------
1. Setup Theos on an iOS device 
2. Try to compile a project

What happens?
-------------
Various incompatible architecture errors 

What were you expecting to happen?
----------------------------------
The project to compile without error(s).

Any logs, error output, etc?
----------------------------
...

Any other comments?
-------------------
What needs updating: 

- iOS Toolchain package
- Theos Dependencies package

My Thoughts:

- Since all required packages will likely remain available on Procursus or Bingner/Elucubratus moving forward, it may be worth creating a new Theos Dependencies package, labeling it "Legacy," and giving it the current package's Depends: and Description: fields. This would allow allow older versions to remain supported while not hindering future versions. Unrelated note: the description is missing "unzip" from the list of packages installed. 
 
- Since the current Theos Dependencies package will no longer be hindered by legacy compatibility, a new, generic toolchain (that would also likely want to be on BigBoss) should be published that depends on clang, ld64, odcctools, make, and ldid. 

- Since the now "legacy" version supports up to iOS 11 and Procursus is iOS 12+, the Theos Dependencies package and this new toolchain should be installed if firmware >= 12. It may also be worth adding firmware << 12 to the legacy package to ensure that it isn't installed on versions that it's incompatible with. 

- The current Theos Dependencies package should drop CoolStar's toolchain and Perl from its Depends field in favor of the new toolchain and generic Perl. 

- The current Theos Dependencies package's Description should be updated to list https://apt.bingner.com/ and https://apt.procurs.us/ instead of CoolStar's public repo and Sam's other repo. It may also be worth adding a note about bootstrap compatibility (e.g., that 'x' repo should only be installed on 'y and z jailbreaks') so that there's no confusion/accidental installations from incompatible bootstraps.

- This doesn't really make a difference, but org.thebigboss.cacerts can be changed to cacerts for the current Theos Dependencies package since it's available (and preinstalled) by both modern bootstraps. 

What versions of software are you using?
----------------------------------------
**Operating System:**
iOS 14.3

**Toolchain Version:**
N/A

**SDK Version:**
N/A

Diatrus:
My solution to this is going to be offering a Theos Dependencies package on Procursus.

[...]

Issue #307, opened 2018-03-27
Implement tests


kirb:
Tests that actually pass would be pretty nice…

* [ ] Create theos/tests repo and dump in known good (builds out-of-the-box) TypeStatus code
* [ ] Have Theos, Logos, dm.pl CI wget this repo and run its test script (copied from our current extras/ci/ scripts)
* [ ] Split into smaller tests via Theos subprojects
* [ ] Have Theos run the entire test suite; Logos only Logos tests in a logos subproject; dm.pl only its tests in a dmpl subproject
* [ ] Implement tests for NIC including confirming all built in templates work, plus some extra unit test templates

Issue #189, opened 2016-08-10
Dependency installer script


uroboro:
We could provide a shell script that would install the required tools to use Theos. Something like [Torch](http://torch.ch) has with its [install-deps](https://github.com/torch/distro/blob/master/install-deps).

supermamon:
https://github.com/supermamon/install-theos/blob/master/install-deps
Very crude? Too fancy? Forgive my bash skills. Also, I don't have a mac to test it on mac.

kirb:
Done by #657 :tada:

PR #657, opened 2022-07-07
Create Theos installer ('install-theos')

L1ghtmann:
- Creates an install script for Theos that supports macOS, iOS, and Linux (Debian and Arch)

Yup, #189 

N/A

Any other comments?
-------------------
This can be tested without cloning this fork with
`bash -c "$(curl -fsSL https://raw.githubusercontent.com/L1ghtmann/theos/tree/pr8/bin/install-theos)"`

A handful of things:

1) **Testing**
- I tested this on each of the platforms listed below and confirmed it and Theos both work as expected, but this should be tested on each at least once more to confirm. 

- I don't own a legacy device, so that is currently **untested** though I'd imagine it'll work fine as it has the dependencies package and the rest of the process has been confirmed working.

2) **Things to note**
- On Unc0ver: 4 packages (ld64, libtapi10, odcctools, and clang-10) don't seem to install properly. Once installed, they need to be manually 'upgraded' in a package manager. Not sure what package managers are doing differently as the 'upgrades' don't appear after an `apt install -f`, `apt update`, or `dpkg --configure -a`. This can likely be resolved by targeting a meta package instead of installing each dependency individually as the script does now (would require the theos dependencies package to be updated -- see #616).

- On Checkra1n: The script can't be run from /var (no exec perms) and needs to be done from a directory where mobile has ownership (since sdks's master.zip is created temporarily). This could potentially be resolved by curl'ing the sdk zip somewhere else (like a tmp directory). 

3) **Possible additions**
- I played around with Theos on both Fedora (WSL) and FreeBSD (VM) and it doesn't work on either currently (both have toolchain issues and FreeBSD needs additional changes).

- [Fedora](https://github.com/SocMinarch/fedora-wsl): libtinfo (as provided by `ncurses-compat-libs`) is missing version info and compiling it from source with `--with-versioned-syms` didn't resolve the issue. This could be resolved by either figuring out the version thing or by compiling a toolchain from Bingner's source specifically for the platform. 

- FreeBSD: requires changes to the scripts to avoid /bin/bash (see #635) and a toolchain (cc: @CRKatri -- do 
you happen to have a toolchain compiled already?)

- Cygwin has been deprecated, but is it worth adding a legacy flag (or will platform detection suffice?) to the script that supports installing on that platform for completeness? 

- Homebrew has an uninstall script as well as an install script. Is this something we should consider?

4) **TODO**
- Assuming this looks good to you guys (generally speaking), I can work on implementing the necessary changes to support CI so that we can get that up and running again (re: #307).

- This will also allow the installation docs to be simplified greatly -- will only need to mention prerequisites for each platform to run the script and additional things that this script can't install but will prompt for (e.g., Xcode on macOS, repos on iOS, curl on Linux, etc).

5) **Thoughts?**
- The prompts (content and color) are just what seemed right to me at the time. If you'd like any changes in that department, just say the word!

- Though it goes without saying, I'm all ears for changes to the code (functional and/or stylistic).

Where has this been tested?
---------------------------
**Operating System:** …

- iOS -- 14.3 -- Taurine, Unc0ver, Checkra1n
- macOS -- Catalina (VM)
- Linux -- Debian and [Arch](https://github.com/yuk7/ArchWSL) (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


CRKatri:
> * FreeBSD: requires changes to the scripts to avoid /bin/bash (see [Add FreeBSD host platform and use /usr/bin/env in shebangs #635](https://github.com/theos/theos/pull/635)) and a toolchain ( -- do  you happen to have a toolchain compiled already?)

nope, I just use my Linux toolchain with the linuxulator.

L1ghtmann:
@CRKatri Oh, cool! 

Booted my vm up again to give that a try and it works like a charm. Only remaining issue seems to be that none of the targets ($THEOS/makefiles/targets/FreeBSD/*.mk) are configured for FreeBSD, so it can't build anything.

L1ghtmann:
Little update: somehow I overlooked the actual error on Fedora ... the lack of libtinfo version info wasn't critical, so Fedora works now!

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+	if ! [[ -d "$XCODE" && "$(ls -A "$XCODE")" ]]; then
+		error "Xcode is mandatory. The Command Line Tools package isn't sufficient for Theos to work. Xcode provides toolchains for all Apple platforms."
+		exit 3
+	fi
+
+	# Check for Homebrew
+	update "Preparing to install dependencies..."
+	if ! command -v brew &> /dev/null; then
+		read -p "Homebrew, which provides tools Theos depends on, is not installed. Would you like to have it installed for you? [y/n]" hbrew
+		if [[ "$hbrew" == "y" || "$hbrew" == "yes" || "$hbrew" == "Y" || "$hbrew" == "Yes" ]]; then
```
kirb, line 107: Could we make a `bool` function for this? Like Theos’s `__theos_bool`.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+		update "\$THEOS has not been set. Setting now..."
+
+		if [[ "$SHELL_ENV" == "unknown" ]]; then
+			error "Current shell ($CSHELL) is unsupported by this installer. Please set the THEOS environment variable to '~/theos' manually before proceeding."
+			exit 7
+		fi
+
+		# Set $THEOS
+		echo "export THEOS=~/theos" >> "$SHELL_ENV"
+		source "$SHELL_ENV"
```
kirb, line 132: Be careful doing this because it could execute a blocking command. For instance in my setup I have zshrc run `tmux attach` if connected via ssh.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+		else
+			error "Please install the repos mentioned above before proceeding."
+			exit 4
+		fi
+	else
+		# Dependencies pkg is outdated, so need to cater install for Procursus vs Elucubratus
+		if [[ -f "/.procursus_strapped" ]]; then
+			read -p "Do you have 'https://apt.procursus.us' installed in the sources of your jailbreak's primary package manager? [y/n]" ready
+			if [[ "${ready,,}" == "y" || "${ready,,}" == "yes" ]]; then
+				sudo apt install ca-certificates clang coreutils curl dpkg git grep ldid make odcctools perl plutil rsync unzip xz-utils -y --allow-unauthenticated --allow-downgrades && update "Dependencies have been successfully installed!" || (error "Dependency install command seems to have encountered an error. Please see the log above." && exit 5)
```
kirb: Pro has an org.theos.dependencies package so we can install that there.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+		error "Xcode is mandatory. The Command Line Tools package isn't sufficient for Theos to work. Xcode provides toolchains for all Apple platforms."
+		exit 3
+	fi
+
+	# Check for Homebrew
+	update "Preparing to install dependencies..."
+	if ! command -v brew &> /dev/null; then
+		read -p "Homebrew, which provides tools Theos depends on, is not installed. Would you like to have it installed for you? [y/n]" hbrew
+		if [[ "$hbrew" == "y" || "$hbrew" == "yes" || "$hbrew" == "Y" || "$hbrew" == "Yes" ]]; then
+			bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)" && update "Homebrew has been successfully installed!" || (error "Homebrew install command seems to have encountered an error. Please see the log above." && exit 5)
```
kirb: For readability wrap &&s and ||s onto new lines. Like:

```bash
bash -c "$(curl …)" \
	&& update "Homebrew installed" \
	|| (error "Homebrew failed"; exit 5)
```

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+	if [[ -d "$THEOS/toolchain/linux/iphone/" && "$(ls -A "$THEOS/toolchain/linux/iphone")" ]]; then
+		update "A toolchain appears to already be installed."
+	else
+		while true; do
+			update "A toolchain does not appear to be installed."
+			read -p "Would you like your toolchain to support Swift (larger toolchain size) or not (smaller toolchain size)? [y/n]" stoolchain
+			if [[ "${stoolchain,,}" == "y" || "${stoolchain,,}" == "yes" ]]; then
+				read -p "You chose to install the Swift-compatible toolchain. Is this correct? [y/n]" confirm
+				if [[ "${confirm,,}" == "y" || "${confirm,,}" == "yes" ]]; then
+					if [[ "$DISTRO" == "debian" ]]; then
```
kirb, line 355: Use a switch:

```bash
case $DISTRO
debian) sudo apt install zstd -y ;;
arch)   sudo pacman -S --needed --noconfirm zstd ;;
# etc…
*)      # fallback else case if needed
esac
```

Can be multiline if needed of course, just needs to end in `;;`.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+		git clone --recursive https://github.com/theos/theos.git $THEOS && update "Git clone of Theos was successful!" || (error "Theos git clone command seems to have encountered an error. Please see the log above." && exit 9)
+	fi
+
+	# Get patched sdks
+	update "Checking for patched sdks..."
+	if [[ -d "$THEOS/sdks/" && "$(ls -A "$THEOS/sdks/" | grep sdk)" ]]; then
+		update "SDKs appear to already be installed."
+	else
+		update "SDKS do not appear to be installed. Installing now..."
+		curl -LO https://github.com/theos/sdks/archive/master.zip
```
kirb: Since we have quite a few places where we curl and unzip/untar something, could we centralise this to a function? May as well change to `master.tar.gz` so that function can just assume it needs to use tar.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+			exit 9
+		fi
+	fi
+}
+
+
+# Determine platform and start work
+if [[ "$PLATFORM" == "Darwin" ]]; then
+	special "Starting install..."
+	if [[ "$(uname -m)" =~ "iP" ]]; then
```
kirb, line 519: Could we change this to test based on `xcode-select` existence? That’s how Theos tests for macOS.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+	XCODE="/Applications/Xcode.app/"
+	if ! [[ -d "$XCODE" && "$(ls -A "$XCODE")" ]]; then
+		error "Xcode is mandatory. The Command Line Tools package isn't sufficient for Theos to work. Xcode provides toolchains for all Apple platforms."
+		exit 3
+	fi
+
+	# Check for Homebrew
+	update "Preparing to install dependencies..."
+	if ! command -v brew &> /dev/null; then
+		read -p "Homebrew, which provides tools Theos depends on, is not installed. Would you like to have it installed for you? [y/n]" hbrew
```
kirb, line 106: To be really annoying (sorry 😛) could this support port and apt (Procursus) as well?

kirb (changes requested):
This is incredible work. Still pretty busy right now but I’ll review this properly as soon as I can. Just giving you some very high-level suggestions from skimming through first…

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+	if [[ $KERNEL_VER < 1800 ]]; then
+		OLD=1
+	fi
+
+	# Dependencies
+	update "Preparing to install dependencies. Please enter your password if prompted:"
+	if [[ $OLD == 1 ]]; then
+		read -p "Do you have 'https://coolstar.org/publicrepo/', 'https://repo.bingner.com/', and 'http://apt.thebigboss.org/repofiles/cydia/' installed in the sources of your jailbreak's primary package manager? [y/n]" ready
+		if [[ "${ready,,}" == "y" || "${ready,,}" == "yes" ]]; then
+			sudo apt install org.theos.dependencies -y --allow-unauthenticated --allow-downgrades && update "Dependencies have been successfully installed!" || (error "Dependency install command seems to have encountered an error. Please see the log above." && exit 5)
```
kirb: `apt-get` would be a safer bet for legacy, I think we didn’t get `apt` on Telesphoreo.

kirb (changes requested):
Just spotted another thing

bin/install-theos
```diff
@@ -0,0 +1,532 @@
...
+					if [[ "$DISTRO" == "debian" ]]; then
+						sudo apt install zstd -y
+					elif [[ "$DISTRO" == "arch" ]]; then
+						sudo pacman -S --needed --noconfirm zstd
+					elif [[ "$DISTRO" == "redhat" ]]; then
+						sudo dnf install -y zstd
+					elif [[ "$DISTRO" == "suse" ]]; then
+						sudo zypper install -y zstd
+					fi
+					curl -LO https://github.com/CRKatri/llvm-project/releases/download/swift-5.3.2-RELEASE/swift-5.3.2-RELEASE-ubuntu20.04.tar.zst
```
kabiroberai, line 364: Is this still the latest Swift toolchain that we have on Linux?

bin/install-theos
```diff
@@ -0,0 +1,532 @@
...
+	fi
+
+	# Check desire for Swift support
+	update "Checking desire for Swift support..."
+	read -p "Would you like to be able to work with Swift? If so, an additional package will need to be installed. [y/n]" confirm
+	if [[ "${confirm,,}" == "y" || "${confirm,,}" == "yes" ]]; then
+		if [[ -f "/.procursus_strapped" ]]; then
+			sudo apt install swift -y --allow-unauthenticated --allow-downgrades && update "Additional Swift package installation successful!" || (error "Additional Swift package install command seems to have encountered an error. Please see the log above." && exit 3)
+		else
+			sudo apt install com.kabiroberai.swift-toolchain -y --allow-unauthenticated --allow-downgrades && update "Additional Swift package installation successful!" || (error "Additional Swift package install command seems to have encountered an error. Please see the log above." && exit 3)
```
kabiroberai, line 199: tbh I'm not sure how useful this is anymore since Swift 4 (which iirc is what this toolchain has) is suuuper old

bin/install-theos
```diff
@@ -0,0 +1,532 @@
...
+			exit 8
+		fi
+	fi
+}
+
+darwin() {
+	# Check for Xcode
+	XCODE="/Applications/Xcode.app/"
+	if ! [[ -d "$XCODE" && "$(ls -A "$XCODE")" ]]; then
+		error "Xcode is mandatory. The Command Line Tools package isn't sufficient for Theos to work. Xcode provides toolchains for all Apple platforms."
```
kabiroberai, line 99: Let's link to <https://developer.apple.com/download/applications/> to encourage users to get Xcode from developer.apple.com instead of the MAS.

In the future, it would be really cool if we could automatically install and use [xcodes](https://github.com/RobotsAndPencils/xcodes) with `--experimental-unxip` + aria2 support.

kabiroberai (commented):
Thanks a ton for this PR, I can definitely see it being incredibly useful for newer devs! No blocking requests on my end, just a few questions and comments.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+fi
+
+
+# The work
+darwin() {
+	common "Platform: Darwin"
+
+	# Check for Xcode
+	XCODE="/Applications/Xcode.app/"
+	if ! [[ -d "$XCODE" && "$(ls -A "$XCODE")" ]]; then
```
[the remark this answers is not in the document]
L1ghtmann: Doesn't list "." and ".."

bin/install-theos
L1ghtmann: Ooo, good idea!

bin/install-theos
L1ghtmann: I believe so, yeah.

bin/install-theos
L1ghtmann: I don't think there's an alternative available for Eluc. Is something better than nothing here or is it just plain not useful at all since it's outdated?

bin/install-theos
L1ghtmann: Yup. For some reason Proc isn't detected as signed immediately after jailbreaking (can be resolved with an `apt update`), but since that's a rare case it shouldn't be a huge issue (or may be a fluke on my device).

bin/install-theos
L1ghtmann: Switched 'legacy' dep install to apt-get and kept the others as apt.

bin/install-theos
absidue: @L1ghtmann Probably best to apt update first anyway as you will want to install the latest versions of the dependencies, to get the best experience.

bin/install-theos
L1ghtmann: Went with xcode-select.

bin/install-theos
```diff
@@ -0,0 +1,598 @@
...
+	NONINTERACTIVE=1
+fi
+
+
+# The work
+darwin() {
+	common "Platform: Darwin"
+
+	# Check for Xcode
+	XCODE="/Applications/Xcode.app/"
```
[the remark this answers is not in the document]
L1ghtmann: I had it check the xcode-select path originally and then scrapped it...can't remember why. Anyway, added it back + the suffix check.

bin/install-theos
L1ghtmann: Added the download link to the 'missing Xcode' prompt.

bin/install-theos
L1ghtmann: Done. When all three are missing, however, the installer will only prompt about Homebrew as Procursus for MacOS isn't public-release ready and MacPorts is a bit more involved to both install and remove if desired.

bin/install-theos
L1ghtmann: Is this still something you're interested in? 

The sdk curl has been moved to its own function to avoid duplication for each platform and the only other curls are the Homebrew install script (for those who don't have it installed and wish for it to be) and the non-Darwin toolchains (2xLinux and 2xFreeBSD) which can be its own function if you want to slim that down.

bin/install-theos
CRKatri: > Is this still the latest Swift toolchain that we have on Linux?

I highly encourage you to make a new one, I won't be maintaining that one.

uroboro:
I set up a bunch of VMs with Vagrant and these boxes:

- `ubuntu/kinetic64`
  - It'd be nice to have a `sudo apt update` before installing packages but otherwise installs Theos, and creates and compiles a tweak project.
- `debian/bullseye64`
  - The box is missing the `LC_ALL` env var but otherwise installs Theos, and creates and compiles a tweak project.
- `opensuse/Leap-15.4.x86_64`
  - Installs Theos and creates a tweak project but then fails to compile it with
  ```
  ==> Compiling Tweak.x (armv7)…
  /home/vagrant/theos/sdks/iPhoneOS14.5.sdk/usr/include/module.modulemap:408:11: error: header 'sched.h' not found
     header "sched.h"
            ^
  /home/vagrant/theos/Prefix.pch:18:10: note: submodule of top-level module 'Darwin' implicitly imported here
        #import <AvailabilityVersions.h>
                ^
  /home/vagrant/theos/sdks/iPhoneOS14.5.sdk/usr/include/module.modulemap:408:11: error: header 'sched.h' not found
     header "sched.h"
            ^
  /home/vagrant/theos/Prefix.pch:24:12: note: submodule of top-level module 'Darwin' implicitly imported here
                          #import <TargetConditionals.h>
                                  ^
  2 errors generated.
  make[3]: *** [/home/vagrant/theos/makefiles/instance/rules.mk:262: 
  /home/vagrant/a/.theos/obj/debug/armv7/Tweak.x.d6510a56.o] Error 1
  rm /home/vagrant/a/.theos/obj/debug/armv7/Tweak.x.m
  make[2]: *** [/home/vagrant/theos/makefiles/instance/library.mk:52: 
  /home/vagrant/a/.theos/obj/debug/armv7/A.dylib] Error 2
  make[1]: *** [/home/vagrant/theos/makefiles/instance/library.mk:37: internal-library-all_] 
  Error 2
  make: *** [/home/vagrant/theos/makefiles/master/rules.mk:117: A.all.tweak.variables] Error 2
  ```
  And it's not an armv7 issue as it happens as well if I set ARCHS to arm64.
- `freebsd/FreeBSD-13.1-RELEASE`
  - Got `"==> Current shell (csh) is unsupported by this installer. Please set the THEOS environment variable to '~/theos' manually before proceeding."` We could support csh by writing to `~/.cshrc`
- `fedora/36-cloud-base`
  - The installer causes the shell to exit after printing `"==> Preparing to install dependencies. Please enter your password if prompted:"`
- `archlinux/archlinux`
  - Fails to install `z3-4.9.1-1`
- `alexgleason/elementaryos-freya64`
  - Fails with `E: Unable to locate package libz3-dev`
- `centos/7`
  - `git` is not installed and therefore fails after printing `"==> Theos does not appear to be installed. Cloning now..."`

My `Vagrantfile` is this (after removing all comments):

```ruby
Vagrant.configure("2") do |config|
	config.vm.define "ubuntu" do |ubuntu|
		ubuntu.vm.box = "ubuntu/kinetic64"
	end

	config.vm.define "debian" do |debian|
		debian.vm.box = "debian/bullseye64"
	end

	config.vm.define "opensuse" do |opensuse|
		opensuse.vm.box = "opensuse/Leap-15.4.x86_64"
	end

	config.vm.define "freebsd" do |freebsd|
	 	freebsd.vm.box = "freebsd/FreeBSD-13.1-RELEASE"
	end

	config.vm.define "fedora" do |fedora|
		fedora.vm.box = "fedora/36-cloud-base"
	end

	config.vm.define "archlinux" do |archlinux|
		archlinux.vm.box = "archlinux/archlinux"
	end

	config.vm.define "elementary" do |elementary|
		elementary.vm.box = "alexgleason/elementaryos-freya64"
	end

	config.vm.define "centos" do |centos|
		centos.vm.box = "centos/7"
	end

	config.vbguest.auto_update = false
	config.vbguest.no_remote = true
end
```

I tried getting the latest versions of all of these but Centos' latest release is 2004 so I don't think we need to worry too much about it or other old OSes.

uroboro:
Tried again everything:

- `ubuntu/kinetic64`
  - All good.
- `debian/bullseye64`
  - I don't know what's wrong with the box that keeps saying `bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)` but otherwise works fine.
- `opensuse/Tumbleweed.x86_64`
  - Got these missing packages errors:
    ```
    'libxml2' not found in package names. Trying capabilities.
    'libz3-4_8' not found in package names. Trying capabilities.
    No provider of 'libz3-4_8' found.
    ```
- `opensuse/Leap-15.4.x86_64`
  - No changes.
- `freebsd/FreeBSD-13.1-RELEASE`
  - We can configure the env var with `echo 'setenv THEOS $HOME/theos' >> $HOME/.cshrc` but then bash is actually csh so make fails to do anything.
  - For the make/gmake issue, we can do `echo 'alias make gmake' >> $HOME/.cshrc`
  - I reached `/home/vagrant/theos/makefiles/common.mk:156: *** The "iphone" target is not supported on the "" platform.  Stop.`, this should be fixed with #635 as mentioned in the OP PR text.
- `fedora/36-cloud-base`
  - No changes but I can't find a fedora box that's not cloud (https://app.vagrantup.com/fedora).
- `archlinux/archlinux`
  - No changes.
- `centos/8`
  - Upgraded from 7 but still fails:
    ```
    CentOS Linux 8 - AppStream                      517  B/s |  38  B     00:00    
    Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
    ```

I think it's ready but I'm not sure I agree with the changes in https://github.com/theos/theos/pull/657/commits/20ac0c04bbbe4c492426db0712d78e0094360582 allowing insecure repos by default.

kirb (approved):

PR #665, opened 2022-09-10
Add blank SYSROOT check and use '__exists', '$(_THEOS_TRUE)', and '$(_THEOS_FALSE)' where applicable 

L1ghtmann:
- Adds a check for a blank sysroot (result of the sysroot wildcard check failing for both $THEOS/sdks and XCode's sdk dir)
  - This would fail before the changes here, but the error was somewhat unintuitive to new devs as clang states that the sysroot is "''" when they did in fact specify an sdk, it just doesn't exist in either of the sdk dirs
  - ```clang-10: error: no such sysroot directory: '' [-Werror,-Wmissing-sysroot]```
- Uses __exist and the $(_THEOS_BOOL) statements where applicable 

Nope

N/A

Any other comments?
-------------------
If you'd like anything to be changed or reverted just say the word.

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


makefiles/master/rules.mk
```diff
@@ -46,6 +46,10 @@ ifneq ($(SYSROOT),)
 ifneq ($(call __exists,$(SYSROOT)),$(_THEOS_TRUE))
 	$(ERROR_BEGIN) "Your current SYSROOT, “$(SYSROOT)”, appears to be missing." $(ERROR_END)
 endif
+# SYSROOT existence check failed so
+# it's now blank and we're bailing out
+else
+	$(ERROR_BEGIN) "Your current SYSROOT does not exist." $(ERROR_END)
```
kirb, line 52: Since the error case is that we didn't find any sysroot at all (user didn't set SYSROOT directly, and darwin makefiles didn't find one in Xcode or $THEOS/sdks/), I think this should be reworded to be more clear about the problem and provide a link to the docs for installing an SDK.

kirb (changes requested):

kirb:
Merged with message reworded, thanks!

PR #689, opened 2023-01-07
Remove a retired script from `THEOS_BIN`

L1ghtmann:
- ~~Merges `install.mergeDir`, `install.copyFile`, and `install.exec` into a single `install.sh` script~~
- ~~Explicitly uses `$THEOS_BIN_PATH` in each invocation (just to be safe)~~
- Removes `deb_build_num.sh` as it was retired more than a decade ago in https://github.com/theos/theos/commit/362fafc6c78dc90d92fc2e30e9c1ebed91041af7

I don't think so, no

N/A

Any other comments?
-------------------
I tested this by running `make do` and it seemed to hold up!

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL) 

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


kabiroberai:
I'm in favour of removing `deb_build_num.sh`, and to some extent specifying `$THEOS_BIN_PATH` for the install invocations (though this does remove the ability to override the scripts via the `PATH`).

Wrt the install script files though, is there a compelling reason to merge them into one? I'm in the "if it ain't broke don't fix it" camp in that case, especially because there may be other tools/modules relying on those scripts, and (imo) bundling them increases cognitive overhead. I can be convinced otherwise if other people disagree though.

kirb:
`install.exec` is used in [far too many](https://github.com/search?utf8=%E2%9C%93&q=%22install.exec%22+%22common.mk%22+lang%3AMakefile+&type=code) projects for us to remove it. If we merge them we still need to support these, but then we may as well not do this refactoring if the originals need to be kept around to redirect to the new equivalents anyway.

L1ghtmann:
Oop, that's fair. Worth doing the `deb_build_num.sh` and/or `$THEOS_BIN_PATH` changes or should I just scrap this pr?

kabiroberai:
Let's avoid the `$THEOS_BIN_PATH` change since the templates don't use it. I'm good with removing `deb_build_num.sh` to reduce clutter.

leptos-null (approved):
Seems straight forward, and I've confirmed with [GitHub search](https://github.com/search?q=deb_build_num.sh&type=code) that no one appears to be using this script.

uroboro (approved):

PR #696, opened 2023-01-15
Use 1 for default deb compression

L1ghtmann:
- With https://github.com/theos/dm.pl/commit/4d3e455f632f77bbf500e03c3670b7cd951f546f, compression level zero became equivalent to `cat`, thus resulting in the creation of `data.tar` instead of `data.tar.lzma` 

Not that I'm aware of, no 

N/A

Any other comments?
-------------------
N/A

Where has this been tested?
---------------------------
**Operating System:** …

WSL (Linux)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


kabiroberai:
fyi we intentionally want minimal compression in debug builds to minimize build time as much as possible. Is there an issue with non-compressed `data.tar` files?

L1ghtmann:
Nope, no compression still conforms to the spec, but I figured that the current expected functionality would want to be maintained. If not, I can scrap this.

kirb:
It definitely should be 1, because some file size reduction is better than none. Running lzma briefly for some quick compression saves a lot of time in transferring large debs especially over USB or slow wifi.

kirb (approved):

PR #705, opened 2023-03-04
Fix procursus repo url on darwin-mobile

JordanEJ:
I have just fixed a wrong url in the installer. When trying to install theos on iOS 12+ it askes if you have https://apt.procursus.us repo in on your package manager. I have changed it to ask for https://apt.procurs.us which is the correct url.


PoomSmart (approved):

kirb (approved):
Good catch, thanks for the PR!

PR #710, opened 2023-03-26
Export THEOS_PACKAGE_INSTALL_PREFIX to CFLAGS

opa334:
Exports THEOS_PACKAGE_INSTALL_PREFIX to CFLAGS so that it can be accessed by preprocessor macros


kirb (approved):
Thanks heaps :rocket:

PR #712, opened 2023-03-28
Minor rootless impl adjustments

L1ghtmann:
- Keeps $THEOS_LIBRARY_PATH/iphone/rootless (but none of the contents) in order to avoid missing directory warnings during linking
- Removes the $(or... $THEOS_PACKAGE_SCHEME "rootful" fallback as the corresponding directory doesn't exist and isn't used internally 
- Moves the $THEOS_PACKAGE_SCHEME flags into 'else' statements to keep them and the default flags separate 

I don't think so

N/A

Any other comments?
-------------------
I only briefly checked the flags with messages=yes so apologies if I missed something

Also, do the library search paths need to be updated depending on the scheme as well or is that not necessary?

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


PoomSmart:
@kirb

PR #715, opened 2023-04-01
[deb] Remove /var/jb if it's empty on building package

PoomSmart:

1. You have a tweak/tool/etc. project and all you have there are DEBIAN scripts. No such files to put in the filesystem.
2. You build the project with `THEOS_PACKAGE_SCHEME=rootless`.
3. `/var/jb` directory is created in the output deb file.
4. You install this deb package on a rootless jailbroken device.
5. You install any other rootless package, and the previous package will be wiped out because APT (?) assumes that `/var/jb` is removed.

The changes make sure that no empty `/var/jb` directory will be included in the final deb.

No

This is a message you see in the install log as you install any rootless package that will wipe out the previous one:

`(Noting disappearance of <Package ID>, which has been completely replaced.)`

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


L1ghtmann (approved):

L1ghtmann:
If #732 happens to be merged in its current state prior to this, it may be worth revisiting the solution here. In said pr, I move all of the non-debian staging items into the prefixed path, meaning that for all projects that aren't of null type, /var/jb (or whatever THEOS_PACKAGE_INSTALL_PREFIX is) wouldn't be empty. In that case, it may be worth adding an additional check underneath the package scheme one to check for the contents of the staging dir to ensure it isn't just the debian directory, in which case the package install prefix dir should not be made in the first place. 

I just did a few quick tests and 
```make
ifneq ($(lastword $(subst /, , $(wildcard $(THEOS_STAGING_DIR)/*)))$(words $(wildcard $(THEOS_STAGING_DIR)/*)),DEBIAN1)
``` 
seemed to work alright, though there are definitely other ways to go about checking that the staging dir is "empty."

PoomSmart:
@L1ghtmann With #732 merged, I believe you have a way around this?

PoomSmart:
Superseded by #775

PR #719, opened 2023-04-13
[bin] Require gzip for install-theos

0x8ff:
See https://github.com/theos/theos.dev/pull/34

Where has this been tested?
---------------------------
**Operating System:** iOS 14.2

**Platform:** Odysseyra1n

**Target Platform:** Odysseyra1n

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -250,7 +250,7 @@ darwin_mobile() {
 			read -p "Do you have 'https://apt.bingner.com' installed in the sources of your jailbreak's primary package manager? [y/n]" ready
 			if theos_bool $ready; then
 				sudo apt update --allow-insecure-repositories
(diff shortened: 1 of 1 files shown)


L1ghtmann (changes requested):
Thanks again! Taking a look at the three `dpkg`'s available -- Telesphoreo's, Elucubratus', and Procursus', it seems as though `gzip` is listed as a dependency on all platforms except Procursus, whose `dpkg` does not appear to depend on `gzip`. As such, this change likely belongs in the proceeding 'if' statement's install command as opposed to the one for non-Procursus-based jailbreaks. Alternatively, the dependency could be added directly to the dependencies package which is available [here](https://github.com/ProcursusTeam/Procursus).

PR #723, opened 2023-04-15
Minor cleanup of .xi/.xmi impl

### LINKED ISSUE #682: [Bug]: `xi` and `xmi` files are run through Logos twice, once outside of the preprocessor

1. Make a new project with `nic.pl`
2. Add a `.xi` or `.xmi` file (ie. like the following:)
	
	<details>
		<summary>Sample file</summary>
	
	```objc
	#include <Foundation/Foundation.h>
	#include <substrate.h>
	
	#ifdef CONDITIONALLY_ENABLED
	
	%hookf(void, randomFunc) {
	  return %orig;
	}
	
	#endif
	
	%ctor {
	#ifdef CONDITIONALLY_ENABLED
	    %init(randomFunc = MSFindSymbol(NULL, "_randomFunc"));
	#else
		%init;
	#endif
	}
	```
	</details>

3. Attempt to build with `make`
4. Get an error signifying that the file is being run through Logos without being run through the preprocessor

### What happens?

The file is run through the preprocessor, and then through Logos, as expected. However, after that, Logos is run on the original file for some reason, which fails when preprocessor directives are controlling whether some hooks are used or not.

File is not run through Logos for a second time, and only run through Logos after being run through the preprocessor, as stated on [Logos: File Extensions](https://theos.dev/docs/logos-file-extensions).

With the sample file posted above:

```shell
dhinak@Dhinaks-MacBook-Air testrepo % make
==> Notice: Build may be slow as Theos isn’t using all available CPU cores on this computer. Consider upgrading GNU Make: https://theos.dev/docs/parallel-building
==> Warning: Building for iOS 7.0, but the current toolchain can’t produce arm64e binaries for iOS earlier than 14.0. More information: https://theos.dev/docs/arm64e-deployment
> Making all for tweak TestRepo…
==> Preprocessing Tweak.xmi…
Tweak.xmi:16: error: re-%init of %group _ungrouped, first initialized at Tweak.xmi:14
make[3]: *** [/Users/dhinak/Documents/testrepo/.theos/obj/debug/armv7/Tweak.mii] Error 255
rm /Users/dhinak/Documents/testrepo/.theos/obj/debug/armv7/Tweak.mii
make[2]: *** [/Users/dhinak/Documents/testrepo/.theos/obj/debug/armv7/TestRepo.dylib] Error 2
make[1]: *** [internal-library-all_] Error 2
make: *** [TestRepo.all.tweak.variables] Error 2
```

### Additional context

Commenting out L281/L291 in https://github.com/theos/theos/blob/79815fc71202ccae34cf9a3fb04c81af8864f1b2/makefiles/instance/rules.mk and changing `$*.mi`/`$*.mii` on L286/L296 to `$*.xi.mi`/`$*.xmi.mii` will lead to a successful build, but I'm opening this issue because I don't know if there's any reason for that second Logos run.

macOS

macOS 12.5

Xcode 14.1

iOS 14.5 from theos/sdks

iOS

iOS 13.1.1

### Which device model is it?

iPhone SE

_No response_

L1ghtmann:
- Removes seemingly superfluous call to logos on original .xi/.xmi file post-preprocess and subsequent logos process
- Manually removes the intermediate logos .xi.mi/.xmi.mii files 
  - Note: .pre files still remain in `$(THEOS_OBJ_DIR)`

Resolves #682 

Any other comments?
-------------------
- Changes tested and implemented as described in the original issue by @dhinakg 
- Additional testing done on two known-working .xi/.xmi examples
- Figured I'd pr as the OP never received a response as to their proposed changes

Where has this been tested?
---------------------------
**Operating System:** …

WSL (Linux)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


makefiles/instance/rules.mk
```diff
@@ -292,22 +292,20 @@ $(THEOS_OBJ_DIR)/%.xm.$(_THEOS_OBJ_FILE_TAG).o: %.xm $(THEOS_OBJ_DIR)/%.xm.mm
 $(THEOS_OBJ_DIR)/%.mi: %.xi
 	$(ECHO_NOTHING)mkdir -p $(dir $@)$(ECHO_END)
 	$(ECHO_PREPROCESSING)$(ECHO_UNBUFFERED)$(TARGET_CXX) -x objective-c -E -I"$(call __clean_pwd,$(dir $<))" $(_THEOS_INTERNAL_IFLAGS_C) $(ALL_DEPFLAGS) $(ALL_PFLAGS) $(PREPROCESS_ARCH_FLAGS) $(ALL_OBJCFLAGS) $< > $(THEOS_OBJ_DIR)/$<.pre && $(THEOS_BIN_PATH)/logos.pl $(ALL_LOGOSFLAGS) $(THEOS_OBJ_DIR)/$<.pre > $(THEOS_OBJ_DIR)/$<.mi $(ECHO_END)
-	$(ECHO_NOTHING)$(THEOS_BIN_PATH)/logos.pl $(ALL_LOGOSFLAGS) $< > $@$(ECHO_END)
```
uroboro, line 295: Would we want to clean up the temporary `$(THEOS_OBJ_DIR)/$<.pre`?

I was going to suggest piping to logos.pl but it doesn't take stdin as input (and adding support for it might be complicated as it performs seeks on the input file)

makefiles/instance/rules.mk
```diff
@@ -292,22 +292,20 @@ $(THEOS_OBJ_DIR)/%.xm.$(_THEOS_OBJ_FILE_TAG).o: %.xm $(THEOS_OBJ_DIR)/%.xm.mm
...
 $(THEOS_OBJ_DIR)/%.xi.$(_THEOS_OBJ_FILE_TAG).o: %.xi $(THEOS_OBJ_DIR)/%.mi
 	$(ECHO_NOTHING)mkdir -p $(dir $@)$(ECHO_END)
 	$(ECHO_COMPILING)$(ECHO_UNBUFFERED)$(TARGET_CXX) -x objective-c -c $(ALL_CFLAGS) $(ALL_OBJCFLAGS) $(THEOS_OBJ_DIR)/$<.mi -o $@$(ECHO_END)
-	$(ECHO_NOTHING)rm $(THEOS_OBJ_DIR)/$*.mi$(ECHO_END)
+	$(ECHO_NOTHING)rm $(THEOS_OBJ_DIR)/$*.xi.mi$(ECHO_END)
 
 $(THEOS_OBJ_DIR)/%.mii: %.xmi
 	$(ECHO_NOTHING)mkdir -p $(dir $@)$(ECHO_END)
 	$(ECHO_PREPROCESSING)$(ECHO_UNBUFFERED)$(TARGET_CXX) -x objective-c++ -E -I"$(call __clean_pwd,$(dir $<))" $(_THEOS_INTERNAL_IFLAGS_C) $(ALL_DEPFLAGS) $(ALL_PFLAGS) $(PREPROCESS_ARCH_FLAGS) $(ALL_OBJCFLAGS) $< > $(THEOS_OBJ_DIR)/$<.pre && $(THEOS_BIN_PATH)/logos.pl $(ALL_LOGOSFLAGS) $(THEOS_OBJ_DIR)/$<.pre > $(THEOS_OBJ_DIR)/$<.mii $(ECHO_END)
-	$(ECHO_NOTHING)$(THEOS_BIN_PATH)/logos.pl $(ALL_LOGOSFLAGS) $< > $@$(ECHO_END)
```
uroboro, line 305: Same question as above

uroboro (approved):
Thanks!

makefiles/instance/rules.mk
L1ghtmann: It doesn't look like the .pre's are used anywhere other than the logos calls, so removing them should be alright. Just say the word and I can either add them to the logos intermediate removal command or add a new removal under the rule where they're used.

uroboro: I'd say to add them where they're used (replacing the lines you removed in this PR). I'll try to add support for logos to support stdin as input and then we can pipe through without the intermediate files

L1ghtmann: Done. Also simplified the other removals with `$<` (thanks for that!)

PR #725, opened 2023-04-17
Added .keep for lib/iphone/rootless

MrGcGamer:
This is just GitHub house-keeping `$THEOS/lib/iphone/rootless` already exists, after installation

Don't think so

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


MrGcGamer:
included in #712

MrGcGamer:
As #712 has been closed, without implementing this, I am reopening this PR.

lib/iphone/rootless/.keep
```diff
@@ -0,0 +1 @@
+
```
leptos-null, line 1: nit: this file is 1 byte (line break) instead of empty

leptos-null (changes requested):
Small nit-pick for consistency with other `.keep` files in this repo.

This PR is important for `internal-library-stage_`/`internal-framework-stage_` to work correctly.

leptos-null (approved):
Thanks

uroboro (approved):

leptos-null:
#801 now creates this directory as needed.
I'm merging this PR as well so that the directory is easy to access.
Any number of schemes may exist, so I'm planning to include (via `.keep` files) `lib/{target}/{scheme}` directories in this repositories that have a mirrored directory in https://github.com/theos/lib

PR #737, opened 2023-05-06
Handle @rpaths for libs/frameworks when building for rootless

L1ghtmann:
See title

Don't think so

Any other comments?
-------------------
- Tested via otool on Linux (untested on-device)
- Feel like this makes sense for Theos to handle as THEOS_PACKAGE_SCHEME=rootless already provides the new rpaths, so one might expect it to also convert the install name to @rpath 
  - Opened for discussion in case others feel differently 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


MrGcGamer:
This sounds like a good idea to me and would make sense in my opinion. 👍

leptos-null (approved):

Issue #736, opened 2023-05-04
apt update on Ubuntu may make the theos installer script exit


HMaker:
I tried following the [installation guide](https://theos.dev/docs/installation-linux) but the installer script always exit at `sudo apt update` command. I have some broken apt repositories unrelated to theos, I guess this makes apt update exit the bash script. Weird.

I removed the `sudo apt update` command from the linux installer and it worked.

I am using Ubuntu 20.04.

L1ghtmann:
A potential solution: change the update command to always return true -- `sudo apt update || true`

PR #739, opened 2023-05-08
[bin] remove dnf check-update during dependency install

### LINKED ISSUE #738: [Bug]: Installer script exits if updates are available on Fedora

1. Get a working Fedora install
2. Downgrade a package (any trivial package will do, I just downgraded `alternatives` with `sudo dnf downgrade alternatives`
3. Run the installer script with ` bash -c "$(curl -fsSL https://raw.githubusercontent.com/theos/theos/master/bin/install-theos)"`

### What happens?

The installer script quits after listing the upgradeable packages.

The installer was meant to...install?

Looking at the equivalent Debian code in the installer script, it appears the repositories were meant to be updated, but not any packages, then the installer was meant to proceed.

```shell
fumnanya@fedora: ~ took 7s
➜  bash -c "$(curl -fsSL https://raw.githubusercontent.com/theos/theos/master/bin/install-theos)"
==> Theos Installer: Starting install...

==> Platform: Linux

Last metadata expiration check: 0:10:40 ago on Mon May  8 08:37:14 2023.

alternatives.x86_64                          1.24-1.fc38                           updates

fumnanya@fedora: ~ took 3s
➜
```

### Additional context

_No response_

Linux/WSL

Fedora 38

N/A

N/A

Linux/WSL

N/A

### Which device model is it?

N/A

_No response_

poopsicles:
The `dnf check-update` command in the installer lists the updatable packages and quits the script if there's any.
To fix it, at first I just changed it to `dnf update` which does update them, and it worked.

However, the Debian and Arch behaviours appear to be to "refresh the sources" as opposed to "update the system packages" (`apt update` vs `apt upgrade`, `pacman -Syy` vs `pacman -Syu`) and `dnf update` doesn't follow that intended behaviour.

To make sure the repositories are forcefully updated by dnf, the `--refresh` flag was added to the next line which installs some dependencies.

Closes #738.

None

Any other comments?
-------------------
All that being said, there's a whole thing going on with [Arch partial upgrades](https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported), so maybe `pacman -Syu` would be the way to go for that.

There also appears to be another issue which is similar to this but on Ubuntu (#736), which is weird because from what I remember `apt update` does just update the repos and moves on, but I can't test a fix currently.

Where has this been tested?
---------------------------
**Operating System:** Fedora 38

**Platform:** WSL2

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -318,8 +318,7 @@ linux() {
 				|| (error "Dependency install command seems to have encountered an error. Please see the log above."; exit 3)
 			;;
 		redhat)
-			sudo dnf check-update
-			sudo dnf group install -y "C Development Tools and Libraries" \
+			sudo dnf group install -y "C Development Tools and Libraries" --refresh \
 				&& update "Dependencies have been successfully installed!" \
 				|| (error "Dependency install command seems to have encountered an error. Please see the log above."; exit 3)
 			sudo dnf install -y fakeroot lzma libbsd rsync curl perl zip git libxml2 \


L1ghtmann (approved):
Thanks again! Tested on Fedora 38 and does appear to refresh sources similarly to `apt update`.

PR #746, opened 2023-07-08
Adjust iOS apt update/install flags depending on apt ver

L1ghtmann:
Modifies flags passed to iOS's apt in `install-theos` depending on apt version (1.1 vs 0.6.8 vs older)

I don't believe so

Any other comments?
-------------------
## v1.1
- "Configuration Item: APT::Get::allow-downgrades. Introduced in APT 1.1." [apt-get man](https://manpages.ubuntu.com/manpages/xenial/man8/apt-get.8.html#:~:text=Configuration%20Item%3A%20APT%3A%3AGet,Introduced%20in%20APT%201.1.)
- "Only load unauthenticated data into our parsers when the user explicitly asked for it via --allow-insecure-repositories (Acquire::AllowInsecureRepositories)" [debian/apt](https://github.com/Debian/apt/blob/9ef79abd2f7942c4c5d789fd29b83f493626e440/debian/changelog#L3860)

## v0.6.8
- "Add a config option and corresponding command line option (--allow-unauthenticated) to apt-get, to make buildd operators happy (Closes: #XXXXXX)" [debian/apt](https://github.com/Debian/apt/blob/9ef79abd2f7942c4c5d789fd29b83f493626e440/debian/changelog#L10909)
 
Where has this been tested?
---------------------------
**Operating System:** …

New logic was tested in separate test script - Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


bin/install-theos
```diff
@@ -61,6 +61,34 @@ fi
 
 
 # The work
+ver_compare() {
```
leptos-null, line 64: nit: may you add a comment to this function? e.g.
"compares two version strings in the form of `x.y.z`. returns `1` on failure or if `$1` is less than `$2` otherwise `0`"

leptos-null: Also, I'm not familiar enough with bash, but if `want == have` is the return value well-defined?

leptos-null (approved):
I added a comment to comment a `ver_compare` - after reading the call site, I think its usage is clear. Looks good to me

bin/install-theos
kabiroberai, line 64: Can we assume that dpkg is installed on host systems? If so, we can use `dpkg --compare-versions` instead of hand-rolling our own.

L1ghtmann: Seems --compare-versions was added in dpkg [v1.2.1](https://github.com/oerdnj/dpkg/blob/347f239062f867b6112178338f983803906f5099/debian/changelog#L11104) and  Telesphoreo started at [v1.13.x](http://apt-origin.saurik.com/debs/), so it should be fine?

bin/install-theos
```diff
@@ -220,13 +220,26 @@ darwin_mobile() {
 		exit 3
 	fi
 
+	# Compatbility check
+	APTVER="$(apt-get --version | head -n1 | cut -d' ' -f2)"
+	if [[ $(dpkg --compare-versions $APTVER ge 1.1; echo $?) == 0 ]]; then
```
kabiroberai, line 225: nit: I think this can be simplified to `if dpkg --compare-versions $APTVER ge 1.1; then`

L1ghtmann: Good call, thanks.

kabiroberai (approved):

PR #748, opened 2023-07-08
Mimic frameworks by installing libraries to THEOS_LIBRARY_PATH in internal stage

L1ghtmann:
See title

I don't think so

Any other comments?
-------------------
- Figured this was worth implementing officially as most devs implement their own internal-/after-stage to do just this
- The logic feels a bit messy, so if there is any way to simplify it I'm all ears!

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


uroboro (approved):

leptos-null (approved):
Code looks good (mirrors framework code).
Tested default and rootless, worked well.

PR #754, opened 2023-07-20
CXX Usage Modifications

L1ghtmann:
- Switches relevant file types to CC from CXX
- Defaults to CC for linking and only switches to CXX if OBJ_FILES_TO_LINK contains a ++ file variant
- Skips passing the ++ LD flags if using the CC for linking
- Removes explicit link to libc++ as the CXX *seems* to auto-link it when used as the linker

No

Any other comments?
-------------------
Not tested extensively. Just messages=yes tests to confirm the changes made were effective. Open to discussion.

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


leptos-null (approved):
Looks good to me.

Tested in a project that uses `.x`, `.mm`, and `.c`

Issue #308, opened 2018-03-27
Documentation


kirb:
Our docs are very scattered; almost everything is undocumented! Things that are documented may be on our wiki, dev wiki, DHowett's old docs, or theiostream's [old docs](https://github.com/theiostream/theos-ref). Almost all of those are outdated and missing details.

Docs will be hosted on the GitHub wiki, and mirrored to offline docs shipped with Theos, generated using Jekyll. An Xcode/Dash/Zeal docset would also be nice and would ensure everyone's copy of the docs will stay up-to-date. 

* [ ] Document general concepts (explain the role of Make, Logos; define words such as instance, stage; etc)
* [ ] Document global vars
* [ ] Document instance-specific vars
* [ ] Document rules
* [ ] Document Logos syntax
* [ ] Document NIC syntax
* [ ] Document Logos, NIC, dm.pl use
* [ ] Set up offline docs
* [ ] Set up docsets

uroboro:
Sources to ~~rip~~ take content from, besides the [wiki](https://github.com/theos/theos/wiki) here:

iPhoneDevWiki
- [Theos](http://iphonedevwiki.net/index.php/Theos)
- [Theos/Setup](http://iphonedevwiki.net/index.php/Theos/Setup)
- [Theos/Setup/iOS](http://iphonedevwiki.net/index.php/Theos/Setup/iOS)
- [Theos/Troubleshooting](http://iphonedevwiki.net/index.php/Theos/Troubleshooting)
- [Logos](http://iphonedevwiki.net/index.php/Logos)
- [Logify](http://iphonedevwiki.net/index.php/Logify)
- [NIC](http://iphonedevwiki.net/index.php/NIC)
- [NIC 2 - Templates](http://theos.howett.net/nic/)

DHowett's

- [IPF](http://uv.howett.net/ipf.html)

theiostream

- [theos-ref](https://github.com/theiostream/theos-ref)

uroboro:
Should the components contain their own wiki documenting themselves or should it all fall under the umbrella of Theos' wiki? Or have a separate wiki repo containing all documentation (having the documentation itself under version control)?

kirb:
[...]

PR #757, opened 2023-07-22
Add workflow to add/update offline docs

L1ghtmann:
See title

No, but it will tick another item off of #308

Any other comments?
-------------------
Tested on [separate branch](https://github.com/L1ghtmann/theos/tree/my-docs) and worked like a charm. Would be nice if we could only run the workflow if theos/theos.dev is updated, but not sure that's possible. 

~~Edit: CSS looks to be borked atm~~

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/.github/workflows/offline-docs.yml
+++ b/.github/workflows/offline-docs.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout theos/theos
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
@@ -21,7 +21,7 @@ jobs:
           ruby-version: '2.7.2'
 
       - name: Checkout theos/theos.dev
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: theos/theos.dev
           path: tmp
--- a/.github/workflows/offline-docs.yml
+++ b/.github/workflows/offline-docs.yml
@@ -32,12 +32,12 @@ jobs:
         run: |
           make -C tmp all
           cp -av tmp/_site/docs/ .
-          cp -av tmp/_site/assets/ docs/
+          cp -av tmp/_site/assets/*.css docs/
           rm -rf tmp
 
       - name: Cleanup broken formatting
         run: |
-          sed -i 's/\/assets\//\/docs\/assets\//g' docs/*.html
+          sed -i 's/\/assets\//\/docs\//g' docs/*.html
           sed -i '/<svg.*12.*/,/<\/svg>/d' docs/*.html
           sed -i 's/ Deployed on <a href="https:\/\/netlify.com">Netlify<\/a>.//g' docs/*.html
 
--- a/.github/workflows/offline-docs.yml
+++ b/.github/workflows/offline-docs.yml
@@ -38,6 +38,11 @@ jobs:
 
       - name: Cleanup broken formatting
         run: |
+          # not deploying, so don't need metadata tags or footer note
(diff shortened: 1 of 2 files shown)


.github/workflows/offline-docs.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          sed -i '/<svg.*12.*/,/<\/svg>/d' docs/*.html
+          sed -i 's/ Deployed on <a href="https:\/\/netlify.com">Netlify<\/a>.//g' docs/*.html
+
+      - name: Update docs
+        run: |
+          git config --global user.name "$GITHUB_ACTOR"
+          git config --global user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          git add docs
+          git commit --all -m "[docs] Update offline docs to the latest commit" && git push || echo "No changes were made"
```
leptos-null, line 40: On the line above, you do an explicit add. I think there should not be a `--all` here.

```suggestion
          git commit -m "[docs] Update offline docs to the latest commit" && git push || echo "No changes were made"
```

.github/workflows/offline-docs.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7.2'
+
+      - name: Build docs
+        run: |
+          git clone --depth=1 --recursive https://github.com/theos/theos.dev tmp
```
leptos-null, line 24: nit: use `actions/checkout` for best GitHub integration

```yml
uses: actions/checkout@v3
with:
  repository: theos/theos.dev
  path: tmp
  submodules: recursive

File: .github/workflows/offline-docs.yml [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+          cd tmp
+          make all
```
leptos-null, line 26: consolidate with `make -C tmp`

also fixes what appears to be a bug below: the current working directory is `tmp` and then we `rm -rf tmp`

L1ghtmann: Thanks for catching all of these!

leptos-null (commented):
Overall this code looks good. The workflow will not work though, for two reasons:

1. `secrets.GITHUB_TOKEN` does not have sufficient permissions to write to this repository/ branch, I believe
2. the trigger for this workflow is `on: push: branches: master` however this is unrelated to the event that we're interested in, which is pushes to `master` on `theos/theos.dev`
    - both Lightmann and I originally thought this would work because `update-submodules` would run on organization pushes, however pushes to `theos.dev` would not result in a commit to this repository and therefore would not trigger this workflow

uroboro (approved):

leptos-null:
The first part of my previous comment was not correct: `secrets.GITHUB_TOKEN` does have access to write to this repository

.github/workflows/offline-docs.yml
```diff
@@ -0,0 +1,63 @@
+name: Update Offline Docs
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'master'
```
leptos-null, line 7: In order to respond to pushes on `theos.dev`, we would need a repository dispatch `on` trigger.

We can re-use the organization-wide trigger for this. I think that ideally `theos.dev` would have it's own trigger, but I don't think it's a big enough deal to setup a separate handler for.

```suggestion
  repository_dispatch:
    types: [org_push]
```

leptos-null (approved):
Looks good to me.

Thanks for checking a big item off the to-do list.

PR #759, opened 2023-07-27
Export THEOS_OBJ_DIR and THEOS_STAGING_DIR

Lessica:
Export `THEOS_OBJ_DIR` and `THEOS_STAGING_DIR`. These variables are useful in our `Makefile`.

1. We can use `THEOS_OBJ_DIR` variable to link to other library targets we just built (i.e. internal dependencies).

image

2. We can use `THEOS_STAGING_DIR` variable to do some preprocessing before packaging (e.g. do extra encryption or resource compressing).

image

No

No

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


kabiroberai:
I'm a bit wary of exporting these by default because it might break some multi-Makefile (subproject) builds where the values of these properties differ across makefiles. Could you try to `export THEOS_OBJ_DIR THEOS_STAGING_DIR` in your project makefile instead? These variables should already be accessible from the outside.

kirb:
I’m pretty sure these variables are already passed down, or are at least calculated based on the location of the root project makefile. Are you certain this makes a difference? What were you getting in both variables before adding these lines?

Lessica:
> I’m pretty sure these variables are already passed down, or are at least calculated based on the location of the root project makefile. Are you certain this makes a difference? What were you getting in both variables before adding these lines?

I can't access these variables in bash scripts executed by Makefile.  
Now I solved this problem by `export THEOS_OBJ_DIR THEOS_STAGING_DIR` in my project makefile instead.  Thank you all!

PR #760, opened 2023-07-27
Add flock to gen-commands.pl to avoid race condition

Lessica:
I am not a pro in Perl. But `make commands` sometimes fails to generate a valid json if parallel building is enabled.

No

No, but I found clangd sometimes refused to load compile_commands.json due to syntax errors.

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/gen-commands.pl
+++ b/bin/gen-commands.pl
@@ -1,5 +1,6 @@
 #!/usr/bin/perl
 use JSON::PP;
+use Fcntl ':flock';
 
 $filename = shift;
 $dir = shift;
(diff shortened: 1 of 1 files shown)


kabiroberai (approved):
I suspected this could become an issue but hadn't had the chance to validate/fix it yet. Thanks for tackling this!

Lessica:
> I suspected this could become an issue but hadn't had the chance to validate/fix it yet. Thanks for tackling this!

I found this script might be slow when there're too many project source files.  
It's possible to store build commands somewhere else, in a better structure, and merge them into a JSON at the end.

PR #761, opened 2023-07-29
Don't need to link injector libs if modules are enabled

L1ghtmann:
See title

No

Any other comments?
-------------------
Now that the Orion lib pr has been merged, both libhooker and substrate have module maps and thus auto-link their respective libs on import if modules are enabled 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


leptos-null (approved):
Confirming that in addition to the PR title/ description, this PR also links against `libhooker`, `libblackjack` if needed, which is bug-fix, correct?

If this is intentional, looks good to me.

PR #770, opened 2023-09-09
Installer toolchain updates

L1ghtmann:
- Adds aarch64 toolchain support
- Switches Swift toolchain to Kabir's 
- Updates toolchain deps
- Switches tar commands to curl | tar to remove need for intermediate files
- Fixes dnf 'update' command

No

Any other comments?
-------------------
Swift OS releases > 5.8.1 appear to have a discrepancy with ABI reporting, so this will likely want to be the pinned version utilized until a fix is identified. Cc Leptos

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL); Ubuntu 20.04, Ubuntu 22.04; SUSE Tumbleweed; SUSE Leap; Arch; Fedora 38; Debian

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


bin/install-theos
```diff
@@ -144,11 +144,10 @@ get_sdks() {
 		update "SDKs appear to already be installed."
 	else
 		update "SDKs do not appear to be installed. Installing now..."
-		curl -L https://api.github.com/repos/theos/sdks/tarball -o sdks.tar.gz
 		TMP=$(mktemp -d)
-		tar -xvf sdks.tar.gz --strip=1 -C $TMP
+		curl -sL https://api.github.com/repos/theos/sdks/tarball | tar -xzvf - --strip=1 -C $TMP
```
kabiroberai, line 148: nit: can we `-C` directly into `$THEOS/sdks` and remove the need for `$TMP` entirely?

leptos-null: I think it's helpful because the script only moves the `.sdk` items as opposed to all the scripts and other files in the archive

leptos-null: Just tested - using `tar xf - --strip=1 -C "${THEOS}/sdks/" "*.sdk"` worked well for me

bin/install-theos
```diff
@@ -383,69 +382,58 @@ linux() {
...
+					# toolchain looks for a specific libncurses
+					LATEST_LIBCURSES="$(ls -v /usr/lib/ | grep libncurses.*so | tail -n1)"
+					sudo ln -sf /usr/lib/$LATEST_LIBCURSES /usr/lib/libncurses.so.6
 					;;
 				redhat)
-					sudo dnf install -y z3-libs zstd
-					# libz3-dev equivalent is z3-libs and ...
-					LATEST_LIBZ3="$(ls -v /usr/lib64/ | grep libz3 | tail -n 1)"
-					sudo ln -sf /usr/lib64/$LATEST_LIBZ3 /usr/lib64/libz3.so.4
+					# sudo dnf install -y z3-libs ncurses-libs
```
leptos-null, line 394: why is this all commented out? should we just delete it?

bin/install-theos
```diff
@@ -383,69 +382,58 @@ linux() {
...
 				suse)
-					sudo zypper install -y libncurses5
+					sudo zypper install -y libncurses6
 					;;
 			esac
-			curl -LO https://github.com/L1ghtmann/llvm-project/releases/latest/download/iOSToolchain.tar.xz
-			tar -xvf iOSToolchain.tar.xz -C $THEOS/toolchain/
-			rm iOSToolchain.tar.xz
+			if [[ $(uname -m) == aarch64 || $(uname -m) == x86_64 ]]; then
+				curl -sL https://github.com/L1ghtmann/llvm-project/releases/latest/download/iOSToolchain-$(uname -m).tar.xz | tar -xJvf - -C $THEOS/toolchain/
```
leptos-null, line 431: it this the right link? both

```
https://github.com/L1ghtmann/llvm-project/releases/latest/download/iOSToolchain-aarch64.tar.xz
https://github.com/L1ghtmann/llvm-project/releases/latest/download/iOSToolchain-x86_64.tar.xz
```

are not found for me

leptos-null: never, I went to the repo listed and the most recent release just isn't published yet 👍

bin/install-theos
L1ghtmann: They're TBD. The Swift toolchain currently only supports Arch- and Debian-based distros.

bin/install-theos
L1ghtmann: Yup, want to wait for merge before publishing as the filenames are different and will break the currently revision of the installer which looks for the non-suffixed x64 file.

bin/install-theos
leptos-null:
```
tar: Option --wildcards is not supported
```

It appears the flag does not exist in BSD `tar`

bin/install-theos
leptos-null: ohh, makes sense. nice

bin/install-theos
leptos-null: While have less intermediates would be nice, perhaps more performance and even robust, given this discrepancy I feel more comfortable keeping the current `mktemp` and `mv` than trying to support both `tar` commands

L1ghtmann: Fun. How do we feel about a one-liner?
```
tar --version | grep GNU &> /dev/null && <tar cmd w/ --wildcards> || <tar cmd w/o --wildcards>
```

leptos-null: haha, no worries. that `tar --version` is fairly simple. I'm also good with either.

Attaching macOS 13.5.2 `tar` output for reference:

```
bsdtar 3.5.3 - libarchive 3.5.3 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.8
```

L1ghtmann: Resolving as superseded by https://github.com/theos/theos/commit/1cfe85938ee26cac704f2e03f25ab114c5bfaa2f.

L1ghtmann:
This is just about ready. Will need to publish my aarch64 toolchain, which I'm currently in the process of updating, and update Kabir's toolchain download links once the relevant prs land.

Recent changes include Swift RHEL support, the removal of Swift SUSE support for the time being, and the modification of the RHEL repository refresh cmd as the prior revision was non-functional (whoops). Swift toolchain deps should also be up-to-date.

L1ghtmann:
In the interim while I work on a fix for the issues mentioned above, this pr has been modified to only adjust the following:
1) Introduces a new local bash variable `ARCH` to simplify some of the checks
2) Introduces the v2.3.0 Swift toolchain from Kabir to fix incompatibilities with older Clangs and iOS 15+ NSString format's
3) Switches tar commands to curl | tar to remove need for intermediate files
4) Fixes dnf 'update' command
5) Introduces aarch64 toolchains 
- Note: will need to push my release to latest once approved before merge in order for these to be available for download by the installer 
  
Requesting reviews when possible. Thanks!

leptos-null (approved):
Code looks good and simpler than before.
Tested on `ubuntu-latest` GitHub Actions runner.

PR #775, opened 2023-10-07
Use additional temporary stage for package schemes 

L1ghtmann:
What does this implement/fix? Explain your changes.
---------------------------------------------------
- Stages first in $THEOS_STAGING_DIR
- Then creates a tmp stage with the install prefix
- Moves the top-level contents of $THEOS_STAGING_DIR, save for DEBIAN, to this new prefixed-stage 
- Moves the contents of the tmp dir (i.e., the install prefix) back to the main stage

Does this close any currently open issues?
------------------------------------------
Closes #774 
Resolves issue approached in #715 

Any relevant logs, error output, etc?
-------------------------------------

Any other comments?
-------------------
Note: quotes in `foreach` seem to break the logic. Likely want to fix that before merge 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

PR #744, opened 2023-06-25
Respect package scheme when installing frameworks

L1ghtmann:
What does this implement/fix? Explain your changes.
---------------------------------------------------
See title

Does this close any currently open issues?
------------------------------------------
I don't believe so

Any relevant logs, error output, etc?
-------------------------------------

Any other comments?
-------------------
Previously frameworks were unconditionally installed to $THEOS_LIBRARY_PATH, which remains $THEOS/lib despite a scheme change.

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

PR #776, opened 2023-10-14
Configure backend for module THEOS_PACKAGE_SCHEMEs and modularize `rootless` scheme  

L1ghtmann:
- Adds `$THEOS_MODULE_PATH/schemes/`
- Adds said directory to the module include statement
- Abstracts out rootless scheme logic to its own module 
- Ignores said module in gitignore
- exports THEOS_PACKAGE_INSTALL_PREFIX in package.mk
- includes module version of package.mk in package.mk prior to including the normal and module versions of the package-format.mk
- Includes portable stage changes (#775)
- Includes rootless install_name changes (#737)

No

Any other comments?
-------------------
- Tested package structures and load/id cmds briefly with dpkg-deb/otool, respectively
  - Please double check these
- Supersedes the two prs mentioned above
- Would benefit from #712, #744, and #748 
- Enables other schemes, such as RootHide, to be added as easily as, say, theos-jailed without needing to pr/switch branches
- Totally open to changes, did this real quick since I had a bit

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


.gitignore
```diff
@@ -9,4 +9,5 @@ include/*
 lib/*
 templates/*
 mod/*
+!mod/schemes/rootless
```
kabiroberai, line 12: wdyt about making a `vendor/mod` dir?

makefiles/common.mk
```diff
@@ -123,6 +123,9 @@ _THEOS_HAS_STAGING_LAYOUT := $(call __exists,$(THEOS_LAYOUT_DIR))
 endif
 
 _THEOS_LOAD_MODULES := $(sort $(call __schema_var_all,,MODULES) $(THEOS_AUTOLOAD_MODULES))
+ifneq ($(THEOS_PACKAGE_SCHEME),)
+_THEOS_LOAD_MODULES += schemes/$(THEOS_PACKAGE_SCHEME)
+endif
```
kabiroberai, line 128: I wonder whether we should deprecate explicitly setting `THEOS_PACKAGE_SCHEME` in favour of `MODULES = <scheme>`

kabiroberai (commented):
Love the idea of adding schemes via modules — hope to give this a more thorough review when I have the chance but no blockers atm

.gitignore
L1ghtmann: Hadn't thought of that but it definitely makes sense. Only concern would be that we'd have to add a separate line to attempt to include from there as well as the user mod directory.

makefiles/common.mk
L1ghtmann: I debated that as well. My current thought process is that schemes are modules, but modules aren't necessarily schemes, so keeping the associated variables separate continues that trend. Also helps make managing the necessary changes to internal Theos processes (e.g., stage) easier as we know the state THEOS_PACKAGE_SCHEME is either a valid scheme or blank.

L1ghtmann: Actually, now that I think about it, the stage changes I propose rely on `THEOS_PACKAGE_INSTALL_PREFIX`, not `THEOS_PACKAGE_SCHEME`. The only instance where we seem to rely on `THEOS_PACKAGE_SCHEME` internally is when we modify the linker's library search paths.

uroboro (approved):

.gitignore
leptos-null: That seems normal and aligned with other patterns (`include`, `lib`)

vendor/include
leptos-null, line 1: is this change to the `vendor/include` module intentional?

L1ghtmann: It was not. Thanks!

.gitignore
L1ghtmann: Modified `__mod` to check both the vendor and non-vendor mod dirs for the specified package scheme. Removed the `schemes/` prefix for simplicity.

leptos-null (approved):
Looks great to me.

Testing: 

(for which I created a module called `leptos_mod` and placed in `$THEOS/mod`)

- `make package`
- `make THEOS_PACKAGE_SCHEME=rootless package`
- `make MODULES=leptos_mod package`
- `make THEOS_PACKAGE_SCHEME=rootless MODULES=leptos_mod package`
- `make "MODULES = leptos_mod rootless" package`

Inspected `tree` of `.theos/_` and used `otool` to verify build products.

PR #786, opened 2023-12-22
Add `SKIP_CODESIGN`

L1ghtmann:
Adds variable to toggle codesigning

No

Any other comments?
-------------------
Seems there have been times when users have wanted to disable automatic codesigning for one reason or another and have had to resort to setting `TARGET_CODESIGN` to gibberish to get the check to pass without being overridden to a default. Now, setting `SKIP_CODESIGN=1` will skip codesigning for the given instance. Confirmed via `otool`.

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


leptos-null:
Setting `TARGET_CODESIGN` to be blank seems to achieve the same thing. Is this necessary?

L1ghtmann:
Placing this here as it seems this may need further review. The codesign step isn't necessary for all platforms, so `TARGET_CODESIGN` is set in darwin_head.mk instead of the given platform.mk. The check in instance/rules.mk, which is modified here, handles the codesign command itself, where it's blank for non-Darwin platforms that lack `TARGET_CODESIGN`. Blanking `TARGET_CODESIGN` in the Makefile of a project targeting iOS does nothing as darwin_heaad will set it `ldid`. However, passing `TARGET_CODESIGN=` to a make invocation for such a project does correctly blank the var and skip the codesigning step. May be able to avoid the addition proposed here if this logic can be adjusted.

leptos-null (approved):
Sounds good to me

PR #787, opened 2023-12-22
Add `install-sdk` script

L1ghtmann:
Adds script to quickly download an sdk from our sdk repo

No

Any other comments?
-------------------
Believe this was proposed at one point or another. I don't have a particular feeling toward having it or not having it, but figured I'd put something together for discussion 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/install-sdk
+++ b/bin/install-sdk
@@ -3,8 +3,8 @@
 # Error codes + association:
 # 1 - api.github.com request failed
 # 2 - Invalid argument(s)
-# 3 - No $THEOS
-# 4 - Requested SDK DNE
+# 3 - Requested SDK DNE
+# 4 - No $THEOS
 # 5 - Requested SDK already exists
 # 6 - Failed curl
 
@@ -24,10 +24,15 @@ if [[ $# -ne 1 ]]; then
 	exit 2
 fi
 
+if ! [[ "${sdks[*]}" =~ $1 ]]; then
+	echo "ERROR: we do not currently provide $1.sdk!" >&2
+	exit 3
+fi
+
 if [[ -z "$THEOS" ]]; then
 	echo "ERROR: \$THEOS does not exist." >&2
 	echo "Please install Theos from https://theos.dev/docs/installation before proceeding." >&2
-	exit 3
+	exit 4
 elif ! [[ -d "$THEOS/sdks/" ]]; then
 	mkdir -pv "$THEOS/sdks/"
 fi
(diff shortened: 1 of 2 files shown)


bin/install-sdk
```diff
@@ -0,0 +1,54 @@
...
+    mkdir -pv "$THEOS/sdks/"
+fi
+
+url=""
+if [[ $1 == latest ]]; then
+    url=$(echo "$urls" | grep 'iPhoneOS' | sort -V | tail -n1)
+elif [[ $1 == latest-tv ]]; then
+    url=$(echo "$urls" | grep 'AppleTVOS' | sort -V | tail -n1)
+else
+    url=$(echo "$urls" | grep "$1")
```
leptos-null, line 39: this might end up with multiple lines depending what `$1` is

leptos-null:
Discord conversation for reference: https://discord.com/channels/811490080278839327/811490080871022654/1182823375832371240

bin/install-sdk
L1ghtmann: `$1` is the full sdk name (e.g., iPhoneOS16.5), so I don't believe that will match more than one result

leptos-null: We _expect_ it to be the full SDK name, but only as implied by the sample usage.

`${THEOS}/bin/install-sdk 1` for example errors out.

L1ghtmann: Whoops! Good call. Should be handled now

leptos-null: On my machine, the latest check incorrectly passes if `$1` is a substring. 
For example, `bin/install-sdk 11.4` errors out at `curl` because `url` is `AppleTVOS11.4\niPhoneOS11.4`

I like this StackOverflow answer on checking for array membership: https://stackoverflow.com/a/47541882

```sh
printf '%s\0' "${sdks[@]}" | grep -F -x -z -- "${1}"
```

bin/install-sdk
```diff
@@ -0,0 +1,55 @@
...
+set -e
+
+urls=$(curl -s https://api.github.com/repos/theos/sdks/releases/latest | grep download_url | sed 's/.*: "\(.*\)"/\1/')
+if [[ -z "$urls" ]]; then
+	echo "ERROR: api.github.com request failed?!" >&2
+	exit 1
+fi
+
+sdks=( $(echo "$urls" | sed -e 's/.*\///' -e 's/\.sdk.*//' | tr '\n' ' ') )
+sdks+="latest latest-tv"
```
leptos-null, line 20: Seems like this is behaving differently on different versions of `bash`.

With `GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)` I'm seeing this append result in `AppleTVOS10.2latest latest-tv AppleTVOS11.4 AppleTVOS12.4 ...` (i.e. the string is being appended to the first element of the array).

It looks like this should be

```suggestion
sdks+=( "latest" "latest-tv" )
```

L1ghtmann: Thanks for catching this!

bin/install-sdk
L1ghtmann: Hm, seems like the previous check was for any instance of the str and not an exact match. This works much better as it matches exact strs, thanks!

leptos-null (approved):
Thanks for all your work on this!

bin/install-sdk
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+# Error codes + association:
+# 1 - api.github.com request failed
+# 2 - Invalid argument(s)
+# 3 - Requested SDK DNE
```
PoomSmart, line 6: DNE?

L1ghtmann: "Does not exist"

TheGithubJoshua:
+1 for this

uroboro (approved):

PR #796, opened 2024-03-24
Use oldabi libroot file when compiling with clang < 12.0.0

opa334:
RIght now compilation for oldabi fails due to libroot only having a new ABI file.
Initially I tried using a patched lipo to combine both old ABI and new ABI slices into the same .a, this however does not work.
Now there are two .a files, one with an old ABI arm64e slice and one with a new one and we use the correct one based on the clang version being used when compiling.

Depends on https://github.com/theos/lib/pull/23 being merged first.

No

Where has this been tested?
---------------------------
**Operating System:** macOS

**Target Platform:** iOS


MrGcGamer (commented):
Tested and looks good to me.

PoomSmart:
LGTM but I had my merge permission removed long ago. 😛

PoomSmart (approved):

L1ghtmann (approved):
Though this is not likely to be the final implementation of `libroot`, approving as a means of fixing oldabi builds.

PR #802, opened 2024-05-19
Only link libroot automatically for iphone target

leptos-null:

- On the base branch, theos always tries to link against `libroot`. `libroot` is only compiled for iOS, so this is an error when targeting any other platform.
  - Solutions provide by @L1ghtmann 

No

Where has this been tested?
---------------------------
**Operating System:** macOS

**Platform:** Darwin

**Target Platform:** iOS, macOS

**Toolchain Version:** `Apple clang version 15.0.0 (clang-1500.3.9.4)`

**SDK Version:** iOS 17.4, macOS 14,4


MrGcGamer:
Seems to be having issues when building for rootful with OldABI in my test case
https://discord.com/channels/811490080278839327/811494949849661490/1243463715194404865

L1ghtmann (changes requested):
Turns out that `-hidden-l` was added as an OSX exclusive in ld64 609, so this impl will have to be modified to support older toolchains. `-Wl,--exclude-libs,` was also proposed as an alternative but found to be a Linux exclusive.

MrGcGamer:
So is there a plan regarding the uselessly exported libroot symbols for every project which uses it?

leptos-null:
> So is there a plan regarding the exported libroot symbols for every project which uses it?

Yes. The plan is that the changes will be in a separate PR

L1ghtmann (approved):
Thanks for putting this together!

MrGcGamer:
> > So is there a plan regarding the exported libroot symbols for every project which uses it?
> 
> Yes. The plan is that the changes will be in a separate PR

For reference: https://github.com/theos/lib/pull/25

PR #807, opened 2024-06-08
Simplify and fix WSL v1 check

L1ghtmann:
- WSL can be identified by the "[Mm]icrosoft" substring in the kernel release
  - Previously, this was checked via a lengthy two-part regex
    - Part one worked as expected but part two, the release version, would work for v2 and not v1
      - To remedy these, simpler checks are now used and the wsl v1 check was switched to a ! v2 check

No

Any other comments?
-------------------
Issue was brought to our attention through Discord

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL -- v1 and v2)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -361,9 +361,9 @@ linux() {
 
 	# Check for WSL
 	update "Checking for WSL..."
-	if [[ $(uname -r | sed -n 's/.*\( *Microsoft *\).*/\L\1/ip') == microsoft ]]; then
-		VERSION=$(uname -r | sed 's/.*\([[:digit:]]\)[[:space:]]*/\1/')
-		if [[ $VERSION -eq 1 ]]; then
+	rel="$(uname -r)"
+	if [[ ${rel,,} =~ microsoft ]]; then
+		if ! [[ $rel =~ WSL2 ]]; then
 			update "WSL1! Need to fix fakeroot..."
 			sudo update-alternatives --set fakeroot /usr/bin/fakeroot-tcp \
 				&& update "fakeroot fixed!" \


bin/install-theos
```diff
@@ -361,9 +361,9 @@ linux() {
 
 	# Check for WSL
 	update "Checking for WSL..."
-	if [[ $(uname -r | sed -n 's/.*\( *Microsoft *\).*/\L\1/ip') == microsoft ]]; then
-		VERSION=$(uname -r | sed 's/.*\([[:digit:]]\)[[:space:]]*/\1/')
-		if [[ $VERSION -eq 1 ]]; then
+	rel="$(uname -r)"
+	if [[ ${rel,,} =~ microsoft ]]; then
```
leptos-null, line 365: What does `${rel,,}` do? This doesn't work on macOS bash (this code is in the `linux` block, so if there's an incompatibility, I think that's ok).

```
$ bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)
Copyright (C) 2007 Free Software Foundation, Inc.
$ rel="$(uname -r)"
$ if [[ ${rel,,} =~ microsoft ]]; then
    echo true
  fi
bash: ${rel,,}: bad substitution
```

L1ghtmann: Sets the string to lowercase. And, yup, is a GNU bash feature I believe

leptos-null: Okay, thanks. Sounds good to me.

leptos-null (approved):
I'm not familiar with WSL behaviors, but code looks good to me.

PR #809, opened 2024-06-09
Ensure root makefile is set as project makefile

### LINKED ISSUE #794: [Bug]: Can't include other makefile in `Makefile`

1. Make new project with `nic.pl` (I only tried with making a tweak so far)
2. `cd` into the new project directory
3. `echo "ARHCS = arm64" > other_makefile.mk`
4. Open `Makefile` and replace the `ARCHS = ...` line with `include ./other_makefile.mk`
5. Run `make package`

### What happens?

`make package` outputs the following instead of compiling successfully:
```
> Making all for tweak bugrepro…
gmake[1]: *** No rule to make target 'internal-tweak-all'.  Stop.
gmake: *** [/Users/june/theos/makefiles/master/rules.mk:146: bugrepro.all.tweak.variables] Error 2
```

A successful compilation and packaging

_No response_

### Additional context

Making a new project without modifying the Makefile as described above allows me to compile successfully, so this isn't a problem with my machine. Additionally, adding the `include other_makefile.mk` after the `include $(THEOS)/makefiles/common.mk` allows the project to compile successfully, but that unfortunately doesn't fix my problem because I need to use the other Makefile to set variables before including `common.mk`.

As far as I can tell ([according to the gnu make manual](https://www.gnu.org/software/make/manual/html_node/Include.html)), I'm using the include directive successfully. I've also tried to use a fully-qualified path to include (`include $(CURDIR)/other_makefile.mk`), but that results in the same error.

I've tried this with both make 3.81 (packaged with macOS) and 4.4.1 (installed from homebrew) and the issue appears with both of them. The only difference is that 4.4.1 contains the file+line trace, while 3.81 omits it.

macOS

macOS 14.2.1

Xcode 11.7, Xcode 15.2

13.7

iOS

15.8

### Which device model is it?

iPhone 6s

_No response_

L1ghtmann:
- See https://github.com/theos/theos/issues/794#issuecomment-2156632232 for an explanation 

Appears to resolve #794 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/makefiles/common.mk
+++ b/makefiles/common.mk
@@ -47,7 +47,7 @@ endif
 
 ifeq ($(_THEOS_PROJECT_MAKEFILE_NAME),)
 _THEOS_STATIC_MAKEFILE_LIST := $(filter-out $(lastword $(MAKEFILE_LIST)),$(MAKEFILE_LIST))
-export _THEOS_PROJECT_MAKEFILE_NAME := $(notdir $(lastword $(_THEOS_STATIC_MAKEFILE_LIST)))
+export _THEOS_PROJECT_MAKEFILE_NAME := $(notdir $(firstword $(_THEOS_STATIC_MAKEFILE_LIST)))
 endif
 
 ifeq ($(_THEOS_INTERNAL_TRUE_PATH),)


leptos-null (approved):
Sounds good to me.

I tested on

```
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

PR #810, opened 2024-06-09
Ensure single-arch objs have spaces stripped before copying

### LINKED ISSUE #808: [Bug]: Error when ARCHS has a space at the end

Build any project and set the `ARCHS` to be 1 valid architecture value followed by a space character. 
For example `ARCHS = arm64 `

Sample project

- `Makefile`
  ```make
  TARGET := iphone:clang:16.5:14.0
  ARCHS := arm64 
  
  include $(THEOS)/makefiles/common.mk
  
  TOOL_NAME = test
  
  test_FILES = main.c
  test_INSTALL_PATH = /usr/local/bin
  
  include $(THEOS_MAKE_PATH)/tool.mk
  ```

  Note that the `ARCHS` line ends with a space character

  Screenshot of the Makefile in nano

- `main.c` (not important)
  ```c
  int main() {
	  return 0;
  }
  ```

### What happens?

During the build process an error similar to

```
cp: .theos/obj/debug/test.47ba6b93.unsigned is not a directory
```

is emitted.

The build to complete without errors.

```shell
set -o pipefail; (cp -a .theos/obj/debug/arm64 /test ".theos/obj/debug/test.47ba6b93.unsigned")
cp: /tmp/am/.theos/obj/debug/test.47ba6b93.unsigned is not a directory
```

### Additional context

I believe this is caused by 

https://github.com/theos/theos/blob/16f19fb1da4322513a9ad0283c12f418121773d1/makefiles/instance/rules.mk#L503

if `ARCHS` is 1 word, the value is used directly in the path, and `ARCHS` may include a space, even if it's only 1 word. This causes the space to end up "in" the path, which causes `cp` to process the value as two separate parameters.

macOS

14

Apple clang version 15.0.0 (clang-1500.3.9.4)

16.5

iOS

N/A

### Which device model is it?

N/A

_No response_

L1ghtmann:
- Removes any spaces from the TARGET_ARCHS when its count is == 1
  - Ensures a valid UNIX path

Appears to resolve #808 

**before**
```
> Making all for tweak test…
==> Preprocessing Tweak.x…
==> Compiling Tweak.x (arm64)…
==> Linking tweak test (arm64)…
==> Generating debug symbols for test…
warning: no debug symbols in executable (-arch arm64)
cp: target '/home/lightmann/test/.theos/obj/debug/test.dylib.86a9eb47.unsigned': No such file or directory
```

**after**
```
> Making all for tweak test…
==> Preprocessing Tweak.x…
==> Compiling Tweak.x (arm64)…
==> Linking tweak test (arm64)…
==> Generating debug symbols for test…
warning: no debug symbols in executable (-arch arm64)
==> Signing test…
> Making stage for tweak test…
```

Any other comments?
-------------------
I'm not sure if this is the best place to strip the variable, but it seems like it's treated elsewhere such that space(s) are of no consequence 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/makefiles/instance/rules.mk
+++ b/makefiles/instance/rules.mk
@@ -125,7 +125,7 @@ endif
 endif
 
 ifeq ($(IS_NEW_ABI),1)
-ABI_SUFFIX = 
+ABI_SUFFIX =
 else
 ABI_SUFFIX = _oldabi
 endif
@@ -494,7 +494,7 @@ endif
 ifneq ($(words $(TARGET_ARCHS)),1)
 	$(ECHO_MERGING)$(ECHO_UNBUFFERED)$(TARGET_LIPO) $(foreach ARCH,$(TARGET_ARCHS),-arch $(ARCH) $(THEOS_OBJ_DIR)/$(ARCH)/$(1)) -create -output "$$@"$(ECHO_END)
 else
-	$(ECHO_NOTHING)cp -a $(THEOS_OBJ_DIR)/$(TARGET_ARCHS)/$(1) "$$@"$(ECHO_END)
+	$(ECHO_NOTHING)cp -a $(THEOS_OBJ_DIR)/$(strip $(TARGET_ARCHS))/$(1) "$$@"$(ECHO_END)
 endif
 
 else


L1ghtmann:
Actually, on second look, the GNU make `strip` built-in states

> Removes leading and trailing whitespace from string and replaces each internal sequence of one or more whitespace characters with a single space. Thus, ‘$(strip a b c )’ results in ‘a b c’.

[source](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html)

so maybe we should strip the variable at assignment?

leptos-null:
> so maybe we should strip the variable at assignment?

`strip` which variable on assignment? This fix works well in my test

L1ghtmann:
`TARGET_ARCHS`

leptos-null:
I see. That makes sense to me.

I kind of like it here, because `cp` (in this case) is what needs to be careful about not having spaces, so writing that safety here I think is more safe.

leptos-null (approved):
Looks good to me

PR #811, opened 2024-06-10
Apply framework search paths to all file types during compilation

L1ghtmann:
- See title

- Don't believe so

Minimal config:
```Makefile
TARGET := iphone:clang:latest:7.0

include $(THEOS)/makefiles/common.mk

TOOL_NAME = test

test_FILES = main.c(pp)
test_CODESIGN_FLAGS = -Sentitlements.plist
test_INSTALL_PATH = /usr/local/bin

include $(THEOS_MAKE_PATH)/tool.mk
```
```c
#include <stdio.h>
#include <substrate.h>
int main(int argc, char *argv[], char *envp[]) {
        printf("Hello world!\n");
        return 0;
}
```
```
==> Compiling main.c (arm64)…
...
In file included from main.c:2:
/home/lightmann/theos/vendor/include/substrate.h:1:10: fatal error: 'CydiaSubstrate/CydiaSubstrate.h' file not found
#include <CydiaSubstrate/CydiaSubstrate.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
as `vendor/include/substrate.h` points to the module header `vendor/lib/CydiaSubstrate/CydiaSubstrate.h`

Any other comments?
-------------------
- This was discussed in the Theos discord's theos-general channel 
  - huge thanks to Psychotipe for bringing this to our attention
- Did some digging and it appears that the vendor -F dirs were passed to OBJCFLAGS and SWIFTFLAGS starting in ['16 if extra frameworks were present](https://github.com/theos/theos/commit/5b70a7316eb93193160d0cb4118984ca607dce21)
- Then in '21, Orion branch [moved them](https://github.com/theos/theos/commit/dd2c8a8f19c68bca2bc0fac7380c020dbbf48333) to compilation irrespective of extra frameworks' presence 
  - That is to say that the assumption appears to have been that module includes would only occur for ObjC/Swift and the issue only arose now as we explicitly rely on a module header with recent changes to the substrate.h shim header 
- This has been tested with substrate for .c, .cpp, .m, .mm, .x, and .xm files but **not for Swift** 

Where has this been tested?
---------------------------
**Operating System:** …

Linux (WSL)

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


leptos-null (approved):
Looks good to me.

Tested Swift with

```swift
import CydiaSubstrate

print("Hello, world!")
```

PR #812, opened 2024-06-11
Update install-theos for legacy devices

### LINKED ISSUE #797: [Bug]: Install on Fresh iOS 8.4.1 EtasonJB

On a fresh iOS 8.4.1 install with EtasonJB (don't install any additional packages) attempt the install instructions for Prior to iOS 12.0.

### What happens?

apt-get, head, cut aren't available so this cannot complete.

apt-get, head, and cut available to complete the install.

_No response_

### Additional context

_No response_

iOS

iOS 8.4.1

N/A

N/A

iOS

iOS 8.4.1

### Which device model is it?

iPad 4

EtasonJB

BJamin99:
add checks for further potentially missing dependencies coreutils and xz-utils.

Potentially closes #797 

No

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating System:** 8.4.1

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -242,6 +242,12 @@ darwin_mobile() {
 		exit 3
 	fi
 
+	# Check for apt-get (not installed by default on some jbs)
+	if ! [[ -x $(command -v apt-get) ]]; then
+		error "Please install 'apt' in your package manager before proceeding with the installation."
+		exit 3
+	fi
+
 	# Compatbility check
 	APTVER="$(apt-get --version | head -n1 | cut -d' ' -f2)"
 	if dpkg --compare-versions $APTVER ge 1.1; then
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -230,6 +230,18 @@ darwin_mobile() {
 		exit 3
 	fi
 
+	# Check for coreutils (not installed by default on some jbs)
+	if ! [[ -x $(command -v head) ]]; then
+		error "Please install 'coreutils' in your package manager before proceeding with the installation."
+		exit 3
+	fi
+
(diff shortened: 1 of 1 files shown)


bin/install-theos
```diff
@@ -230,6 +230,18 @@ darwin_mobile() {
 		exit 3
 	fi
 
+	# Check for coreutils (not installed by default on some jbs)
```
L1ghtmann, line 233: Thanks a ton for looking into this! Did we also want to add apt(-get) here to ensure the later commands proceed?

Spoou:
Good

Spoou (commented):
Good

bin/install-theos
L1ghtmann: Thanks! And, yeah, good point. That should probably be wrapped in a !$LEGACY check and prompt about the lack of Swift support. Would you be comfortable adding that to this pr? If not, no worries.

L1ghtmann (approved):
Looks great. Thanks again for bringing this to our attention and for the fix!

L1ghtmann:
Leaving one final note that while moving the iOS Swift install into the non-legacy section prevents attempted installs on pre-64 devices, where it will fail as the toolchain requires arm64*, it also prevents the installer from installing `swift-toolchain` on arm64 devices that are on iOS < 12. 

One way we can adjust this in the future, if desired, is by adding an addition check for `sysctl.cputype` & `sysctl.cpusubtype`, the former of which appears to be `16777228` for arm64* and the latter of which seems to be `1` for 64 and `2`(?) for 64e.

Issue #819, opened 2024-06-24
"import is a language extension" when compiling a C file with the -pedantic flag


pixelomer:
### What are the steps to reproduce this issue?

Compiling [libunarr-ios](https://github.com/pixelomer/libunarr-ios) with the latest version of Theos reproduces the issue.

### What happens?

A Prefix.pch file is imported during compilation that imports the `<AvailabilityVersions.h>` header. Because `#import` is used, this causes a compilation error when the `-pedantic` flag is used.

### What were you expecting to happen?

I expected the project to compile successfully. Removing the problematic part in Prefix.pch fixes the issue.

### Paste any relevant logs, error output, etc.

```shell
$THEOS/Prefix.pch:18:3: error: #import is a language extension [-Werror,-Wimport-preprocessor-directive-pedantic]
        #import <AvailabilityVersions.h>
        ^
1 error generated.
gmake[3]: ***[$THEOS/makefiles/instance/rules.mk:321: /.../libunarr-ios/.theos/obj/debug/arm64/unarr/common/stream.c.o] Error 1
gmake[3]: *** Waiting for unfinished jobs...
```

### What OS are you seeing the problem on?

macOS

### What OS version does it have installed?

macOS 14.5

### What toolchain and version are you using?

Xcode 15.4

### Which SDK version are you using?

iOS 16.5

### What OS is your client device running?

iOS

### What OS version does it have installed?

irrelevant

### Which device model is it?

irrelevant

### Which jailbreak is it using (if relevant)?

irrelevant

L1ghtmann:
Should be resolved by https://github.com/theos/theos/commit/8a93fabf189ab82f2add7a57ec893dabcf7053c2.

PR #822, opened 2024-07-20
Use include instead of import in Prefix.pch for improved compatibility

leptos-null:

This PR fixes an error when compiling projects with `-pedantic` in `CFLAGS`.

The issue is that `import` is not supported in pedantic mode. To resolve this issue, this PR uses `include` instead of `import`. This is safe because the header being included uses a guard check to avoid being included multiple times: https://github.com/theos/sdks/blob/ca52092676249546f08657d4fc0c8beb26a80510/iPhoneOS16.5.sdk/usr/include/AvailabilityVersions.h#L24 

This is primary difference between `import` and `include`.

This PR only changes this one `import` in `Prefix.pch` because the rest of the file is deprecated.

Yes, #819 

No

Any other comments?
-------------------

Tested by compiling https://github.com/leptos-null/ClassDumpRuntime and https://github.com/pixelomer/libunarr-ios

Where has this been tested?
---------------------------
**Operating System:** macOS

**Platform:** Darwin

**Target Platform:** iOS

**Toolchain Version:** `Apple clang version 15.0.0 (clang-1500.3.9.4)`

**SDK Version:** 17.5

### DIFF
--- a/Prefix.pch
+++ b/Prefix.pch
@@ -15,7 +15,7 @@
 #endif
 
 #if __has_include(<AvailabilityVersions.h>)
-	#import <AvailabilityVersions.h>
+	#include <AvailabilityVersions.h>
 #endif
 
 #if !defined(__IPHONE_14_0) && !defined(THEOS_LEAN_AND_MEAN)


L1ghtmann:
Looks good, thanks!

> This PR only changes this one import in Prefix.pch because the rest of the file is deprecated.

I was under the impression that it was too, but it looks like it's still included for all C projects:
https://github.com/theos/theos/blob/77190f62c4535470eb835a974bffaa17ee19f6bb/makefiles/common.mk#L222

The subsequent imports will only skip if the SDK is old enough that it doesn't define `__IPHONE_14_0` and the user doesn't define the deprecated lean and mean variable. Do we want to transition the others over too? I imagine removing it outright would be non-ideal given that auto-imports are expected behavior for older projects (?)

leptos-null:
I'd prefer not to transition the other ones over- mostly to reduce the change in this file that's included in every compilation.

I agree that we should not remove, I think the (existing) `#if !defined(__IPHONE_14_0)` is an appropriate and sufficient for this.

My other thought is that if a project is using `-pedantic`, it makes sense that they can use `-DTHEOS_LEAN_AND_MEAN` as well.

L1ghtmann:
Fair enough. I edited earlier - think the error is specifically due to import use in C files, so the OBJC check should prevent that from occurring even if pedantic is used.

L1ghtmann (approved):

leptos-null:
I see. That also makes sense.

PR #830, opened 2024-11-11
Update Fedora Install Script

Desyncfy:
The current script does not work on Fedora 41, so I have changed it to do so. The issues lies in trying to install the group by the name rather than the id, which is the (current) functionality. (See more at https://dnf5.readthedocs.io/en/latest/changes_from_dnf4.7.html)

No

```bash
Failed to resolve the transaction:
No match for argument: Development Tools
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
==> Dependency install command seems to have encountered an error. Please see the log above.
```

Any other comments?
-------------------
Nope!

Where has this been tested?
---------------------------
**Operating System:** Fedora 41

**Platform:** Framework Laptop 13 _AMD Ryzen 7040Series_

**Target Platform:** N/A

**Toolchain Version:** N/A

**SDK Version:** N/A

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -354,7 +354,7 @@ linux() {
 				|| (error "Dependency install command seems to have encountered an error. Please see the log above."; exit 3)
 			;;
 		redhat)
-			sudo dnf group install -y "C Development Tools and Libraries" --refresh \
+			sudo dnf group install -y "c-development" --refresh \
 				&& update "Dependencies have been successfully installed!" \
 				|| (error "Dependency install command seems to have encountered an error. Please see the log above."; exit 3)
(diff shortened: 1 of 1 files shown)


L1ghtmann:
Hey, thanks a ton for opening this. Do you know if the id version is backwards compatible?

Desyncfy:
@L1ghtmann Opening a fedora 40 vm to see. It should work, but if it doesn't I can probably write a check to make it backwards compatible

Desyncfy:
Alright. Confirmed it works

L1ghtmann (approved):
Thanks so much for identifying the fix and checking compatibility!

PR #835, opened 2025-01-09
install-theos: Allow setting Swift toolchain choice via THEOS_SWIFT_TOOLCHAIN env var

KD-MM2:
Allow setting Swift toolchain choice via THEOS_SWIFT_TOOLCHAIN env var

Allow non-interactive installation, to use it:
```
$ THEOS_SWIFT_TOOLCHAIN=y bash -c "$(curl -fsSL https://raw.githubusercontent.com/theos/theos/master/bin/
install-theos)"
```

The prompt is asking only when `THEOS_SWIFT_TOOLCHAIN` is not set.

No.

Any other comments?
-------------------

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …


L1ghtmann (commented):
Hey! Thanks for the pr. 

I'm a bit torn as to how to proceed as the original intention with the installer was for it to either be strictly interactive or strictly non-interactive, like Brew's https://github.com/Homebrew/install/blob/master/install.sh. 

It seems like we have a few options if that is insufficient. Namely:
- Add options to the script through `getopts`  
- Allow configuration with env vars 
- For both instances, either keep or remove interactive prompts

Curious to hear others' thoughts. Right now, having not given it a ton of thought, I'm leaning towards keeping things as-is as the install command is simple and universal. Open to changing though if there's a good argument to add to/change that behavior.

leptos-null:
Are we able to use `getopts`? The script is generally run directly from the `curl` result, so I'm not sure about that.

I like your point of using Homebrew as a reference for this. That said, I just read the Homebrew install script, and don't see that they have any "questions" at all, even in interactive mode - it seems to me that the interactive mode is just to show the output differently.

Given this, I'm leaning:
- if a `${FLAG_VAR}` is present in interactive mode, use the flag value, and do not prompt for input
- if a `${FLAG_VAR}` is present in non-interactive mode, use the flag value, otherwise use the default value

(such that `${FLAG_VAR}` is a particular flag we decide, for example `${THEOS_SWIFT_TOOLCHAIN}`)

saltman007web (commented):
merge

L1ghtmann:
Implemented the aforementioned changes. Options now include:
- `INTERACTIVE` (defaults to yes unless user overrides or CI)
- `THEOS_SWIFT_TOOLCHAIN` (defaults to n and only prompts in absence of user override and CI)

These will want to be documented if/when we merge in theos.dev/docs/installation-*

Edit: skipped `hbrew` as it is tentatively being [removed](https://github.com/theos/theos/pull/876)

PR #845, opened 2025-07-22
support standard paths for theos rc

6007135:
Add support for theos rc makefile at standard config file paths.

### DIFF
--- a/makefiles/common.mk
+++ b/makefiles/common.mk
@@ -77,7 +77,11 @@ export THEOS_PROJECT_DIR
 
 export PATH := $(THEOS_BIN_PATH):$(PATH)
 
-ifeq ($(call __exists,$(HOME)/.theosrc),$(_THEOS_TRUE))
+ifeq ($(call __exists,$(XDG_CONFIG_HOME)/theos/rc.mk),$(_THEOS_TRUE))
+-include $(XDG_CONFIG_HOME)/theos/rc.mk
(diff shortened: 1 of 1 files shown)


L1ghtmann (commented):
Hey! Thanks a ton for caring about this and sending in a pr! I remember we discussed this a while back. 

Only comment is it seems it would make historical sense to keep the files w/o extension, like `nicrc`. Meaning `rc.mk` -> `rc`. I am very much open to hearing others' opinions on this though.

6007135:
> Hey! Thanks a ton for caring about this and sending in a pr! I remember we discussed this a while back.
Glad to. And to finally get this merged. :)

> it seems it would make historical sense to keep the files w/o extension
Because of the current naming? Eh. There is certinately precedent for files with extensions. I chose to add it here for file type clarity.

> Does anyone have examples of `.config` file names for other projects?
```
.config/awesome/rc.lua
.config/kitty/kitty.conf
.config/ranger/commands.py
.config/calibre/customize.py.json
.config/OpenSCAD/OpenSCAD.conf
.config/inkscape/preferences.xml
.config/vifm/vifminfo.json
.config/Ladybird/Ladybird.ini
```

> we don't need to repeat the project name in the folder and file name; having an accurate file extension seems like a good practice
Agreed on both points.

I think it's good to have the extension. I suppose I could also check for the path w/o ext if desired. With ext would take precedence.

PR #857, opened 2025-12-04
Update `install-theos` Linux distro determination logic

### LINKED ISSUE #832: [Bug]: Installer script can incorrectly detect Linux distro

1. Get a fresh Fedora computer or VM without theos installed
2. `sudo dnf install apt-utils`
3. Attempt to install theos with the installer script
Issue presents itself

### What happens?

Installer incorrectly recognizes the computer as Debian-based (due to the presence of the `apt` command), attempts to install all dependencies with `apt`, dies.

Installer script should realize that the host machine is actually Red Hat based (Fedora) and install the dependencies with `dnf` instead.

```shell
==> Theos Installer: Starting install...

==> Platform: Linux

[sudo] password for cubeb: 
All packages are up to date.   
Error: Unable to locate package build-essential
Error: Unable to locate package fakeroot
Error: Unable to locate package rsync
Error: Unable to locate package curl
Error: Unable to locate package perl
Error: Unable to locate package zip
Error: Unable to locate package git
Error: Unable to locate package libxml2
==> Dependency install command seems to have encountered an error. Please see the log above.
```
```


### Additional context

Line 325 of `install-theos` (the installer script) performs the check for if the command `apt` exists. I installed `apt-utils` since it provides the `apt-ftparchive` command (useful for messing with debs) but this package also includes `apt` which causes the Line 325 check to mis-identify this computer as Debian. I think there ought to be a more robust method for determining the primary package manager that is meant to be used on the current device than simply seeing which command exists.

OS: Fedora Linux Asahi Remix 40 (KDE Plasma) aarch64 
Host: Apple MacBook Pro (16-inch, M1 Max, 2021) 
Kernel: 6.11.8-400.asahi.fc40.aarch64+16k 


Linux/WSL

Fedora Linux Asahi Remix 40

N/A

N/A

iOS

N/A

### Which device model is it?

N/A

_No response_

### PR DESCRIPTION
**Pull request by L1ghtmann**
Relying now on os-release instead of package manager, which can be muddied by user installs

Previously relied on the package manager to determine the Linux flavor. Now relying on ID and ID_LIKE fields in os-release as more robust 

Resolves #832 

Any other comments?
-------------------
…

Where has this been tested?
---------------------------
**Operating System:** …

**Platform:** …

**Target Platform:** …

**Toolchain Version:** …

**SDK Version:** …

### DIFF
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -322,15 +322,24 @@ darwin_mobile() {
 
 linux() {
 	# Determine distro
+	source /etc/os-release 
 	DISTRO="unknown"
-	if [[ -x $(command -v apt) ]]; then
-		DISTRO="debian"
-	elif [[ -x $(command -v pacman) ]]; then
-		DISTRO="arch"
-	elif [[ -x $(command -v dnf) ]]; then
-		DISTRO="redhat"
-	elif [[ -x $(command -v zypper) ]]; then
-		DISTRO="suse"
+	# First pass for namesake distros 
+	case "$ID" in
+	    debian) DISTRO="debian" ;;
+	    arch) DISTRO="arch" ;;
+	    rhel|fedora) DISTRO="redhat" ;;
+	    suse) DISTRO="suse" ;;
+	esac
+
+	# Second pass for derivatives
+	if [[ "$DISTRO" == "unknown" ]]; then
+	    case "$ID_LIKE" in
+	        *debian*) DISTRO="debian" ;;
+	        *arch*) DISTRO="arch" ;;
+	        *rhel*|*fedora*) DISTRO="redhat" ;;
+	        *suse*) DISTRO="suse" ;;
+	    esac
 	fi
 
 	# Check for sudo (not installed by default on some distros)
--- a/bin/install-theos
+++ b/bin/install-theos
@@ -324,6 +324,7 @@ linux() {
 	# Determine distro
 	source /etc/os-release 
 	DISTRO="unknown"
+	
 	# First pass for namesake distros 
 	case "$ID" in
 	    debian) DISTRO="debian" ;;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by L1ghtmann** [comment]
@6007135 would you mind taking a look at this when you have a moment? I think it should be good-to-go. Thanks!

File: bin/install-theos [hunk 1]
```diff
@@ -322,15 +322,39 @@ darwin_mobile() {
 
 linux() {
 	# Determine distro
+	source /etc/os-release
```
6007135: Per the [docs](https://www.freedesktop.org/software/systemd/man/latest/os-release.html), we should fall back to `/usr/lib/os-release` if `/etc/os-release` is not found. I've never seen it in practice, but it may happen in cases of accidental breakage or very minimal containers or such.
```suggestion
	if [ -e /etc/os-release ]; then
		source /etc/os-release
	elif [ -e /usr/lib/os-release ]; then
		source /usr/lib/os-release
	fi
```

bin/install-theos
```diff
@@ -322,15 +322,39 @@ darwin_mobile() {
...
+	        *rhel*|*fedora*) DISTRO="redhat" ;;
+	        *suse*) DISTRO="suse" ;;
+	    esac
+	fi
+
+	# Third pass for legacy
+	if [[ "$DISTRO" == "unknown" ]]; then
+		# parse for /etc/*-release file(s)
+		rel_files=(/etc/*release /etc/*version)
+		rel=$(cat "${rel_files[@]}" 2>/dev/null | tr '[:upper:]' '[:lower:]')
```
6007135: (Already mentioned in #theos-general, just duplicating here for documentation purposes.)
This should operate on the file names, not the contents: it's the name that's indicative of the distro; the contents (if present) are the version.

L1ghtmann: Thanks again for catching this! I can't remember why I convinced myself the contents mattered....

6007135:
Alright, looks good to me.

PR #869, opened 2026-03-09
fix: codesign binary with `adhoc,linker-signed` flags

### LINKED ISSUE #872: Linux arm64 build
Original dylib working in coruna jailbreak

Mach-O 64-bit arm64e (caps: PAC00) dynamically linked shared library, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|NO_REEXPORTED_DYLIBS>

When I clone coruna repo and build  from this build is crashing springboard and causing respring

Mach-O 64-bit arm64e 00) dynamically linked shared library, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|NO_REEXPORTED_DYLIBS>

Image

claration:
SPTM/TXM devices introduced a bit more limitations with binaries that don't contain a signature. Theos (from what I've seen) seems to compile unsigned dylib files then squishes them with `lipo`, which would inherently also give some mismatched signatures..

Now, after compiling this will codesign the binary with new `adhoc,linker-signed` flags, so hopefully they would run on those devices.

###### Before:

image

###### After making changes:

image
 
Flags modified + `Signature=adhoc` is now present.

No.

Any other comments??
Do note that I hardly know my way around the codebase, and I'm glad to make any changes if neccessary. Also, I have not tested this on any iOS device :)

Where has this been tested?
---------------------------
**Operating System:** macOS


kirb:
A few more things to consider:

1. This patch currently seems to cause the merged binary to be signed twice
2. Projects that override `CODESIGN_FLAGS` will lose the extra flag

The flag probably needs to be moved to `targets/_common/iphone.mk` and wrapped in a version check so we only use it for iOS 17.0+.

claration:
> A few more things to consider:
> 
> 
> 
> 1. This patch currently seems to cause the merged binary to be signed twice
> 
> 2. Projects that override `CODESIGN_FLAGS` will lose the extra flag
> 
> 
> 
> The flag probably needs to be moved to `targets/_common/iphone.mk` and wrapped in a version check so we only use it for iOS 17.0+.

I did notice that it codesigned the binary twice after making a pull request, will fix that when I get the chance.

Is there a reason why for iOS 17.0+? Some tweaks may target lower, or people will try to run these projects not intended for their version but often times they'll end up working anyway..

leptos-null:
> Procursus ldid

I see, thanks!

> the `.dylib` files produced by Theos are squished by `lipo` with no further codesignature changes.

This is not the expected behavior. May you confirm your version of Theos with `git -C "${THEOS}" rev-parse HEAD`? And attach the relevant output from `make messages=yes`?

Here's the output on my machine:

```bash
(printf "\e[0;3%im==> \e[1;39m%s…\e[m\n" 4 "Merging tweak test"); set -o pipefail; (/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo -arch arm64 /tmp/test/.theos/obj/debug/arm64/test.dylib -arch arm64e /tmp/test/.theos/obj/debug/arm64e/test.dylib -create -output "/tmp/test/.theos/obj/debug/test.dylib.47ba6b93.unsigned")
(printf "\e[0;3%im==> \e[1;39m%s…\e[m\n" 4 "Signing test"); set -o pipefail; (CODESIGN_ALLOCATE=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/codesign_allocate  ldid -S "/tmp/test/.theos/obj/debug/test.dylib.47ba6b93.unsigned" && mv "/tmp/test/.theos/obj/debug/test.dylib.47ba6b93.unsigned" "/tmp/test/.theos/obj/debug/test.dylib")
```

claration:
> - This flag crashes the ldid currently installed by the Theos installer on macOS
> - This flag causes the ldid currently installed by the Theos installer on Linux to exit with an error

For `ldid`, perhaps theos should install the Procursus version instead?

> - Binaries signed with this flag may not behave as expected on older jailbreaks/ systems

We can restrict the `-Cadhoc` flags to only be set on rootless environments, there's also essentially a guarantee that these resigned binaries will only run on 15.0+ with this environment. Unfortunately older jailbreaks (10-11~ or below) don't register everything in trustcache so it will end up crashing on execution :/ so I don't think it's a possibility for it to be kept as it currently is.

claration:
Closing – due to Dopamine doing dyld hooking which allows binaries regardless of the flags on TXM enabled devices to launch without any issues. (https://github.com/opa334/Dopamine/commit/e632f52b3e5b9166b64de91e3546306cf1109dbb)
Also, it seems extremely inconvenient for existing theos users to install a new version of `ldid` which support adding these new flags to binaries.

